PR #1, opened 2015-05-27
Added Album member to SimpleTrack struct

roelrymenants:
Hi,

I started using your spotify library. Nice and rather complete!

Started using the CurrentUsersTracks function and noticed that the result didn't contain the album of the tracks, which is returned by the spotify web api. Simply adding the Album to the struct makes it available.

### DIFF
--- a/track.go
+++ b/track.go
@@ -24,7 +24,6 @@ import (
 
 // SimpleTrack contains basic info about a track.
 type SimpleTrack struct {
-	Album   SimpleAlbum    `json:"album"`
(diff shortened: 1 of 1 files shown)


zmb3:
Thanks!

It looks like Album is not in the simplified track object (see here: https://developer.spotify.com/web-api/object-model/#track-object-simplified).

Instead, the album should be added to the `FullTrack` struct.

If you want to make this change I'm happy to accept the PR.

roelrymenants:
Done

PR #2, opened 2015-11-26
Don't escape the query string before calling encode.

bturrubiates:
Hi, 

I found this today and started using it but encountered an issue when trying to search with filters in the query string. I think I've tracked down the issue and my before/after testing seems to indicate that the call to `url.QueryEscape` is the issue. I've only lightly tested it (ran `go test` and tried my example).

I added a unit test that fails without this fix, and passes with the fix. I didn't really know how to name it, or where to place it. I can modify that if you'd like. I used your example of Uptown Funk, but used the filter parameter to search for Bruno Mars since he's a featured artist.

### DIFF
--- a/search.go
+++ b/search.go
@@ -142,7 +142,6 @@ func (c *Client) Search(query string, t SearchType) (*SearchResult, error) {
 // If the client has a valid access token, then the results will only include
 // content playable in the user's country.
 func (c *Client) SearchOpt(query string, t SearchType, opt *Options) (*SearchResult, error) {
-	query = url.QueryEscape(query)
 	v := url.Values{}
 	v.Set("q", query)
 	v.Set("type", t.encode())
--- a/search_test.go
+++ b/search_test.go
@@ -66,6 +66,34 @@ func TestSearchTracks(t *testing.T) {
 	}
 }
 
+func TestSearchTrackWithFilter(t *testing.T) {
+	if os.Getenv("FULLTEST") == "" {
+		t.Skip()
+		return
+	}
+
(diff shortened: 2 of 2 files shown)


zmb3:
Thanks for your PRs!  I will take a closer look at this one after the holiday.

bturrubiates:
Sure, thanks for providing this library :)

zmb3:
Fix looks good.

For the majority of the tests, what I did was run the query using Spotify's [Web API console](https://developer.spotify.com/web-api/console/) and then stuck the output in a file (if if it is small enough, right in the test).  The tests then use a dummy client that returns the response in a particular file or string.

The idea was that this would be faster than having dozens of tests that hit the API, and would still give a reasonable confidence that things were being parsed correctly.

Test's that actually do hit the API (there's only one at the moment) are hidden behind a `FULLTEST` environment variable (see `TestSearchAgainstAPI` in `search_test.go`, for example.

If you have other suggestions for how to test this, I'd love to hear them.  Otherwise, I think we either hide your new test behind the `FULLTEST` environment variable or mock it out like the other tests.  After that I'll gladly merge this.  Thanks again for the contribution.

bturrubiates:
Sure, I can try to mock it out like the other tests. I'll take a crack at that later tonight.

bturrubiates:
Actually, after looking at this again I think it's best to just hide it behind the `FULLTEST` environment variable. If the mocking is used, then it won't fail as expected if something changes in the future since the issue was with the actual request itself rather than the processing of the result.

zmb3:
I agree, the mocked response would not have caught this error.

PR #5, opened 2016-01-19
Fix parameter for adding tracks to playlist: should be "uris"

brandur:
Fixes a bug in the library that causes in adding tracks to a playlist
currently being non-functional. The library passes "urls", but [as
described in the documentation][doc], it should actually be "uris".

[doc]: https://developer.spotify.com/web-api/console/post-playlist-tracks/#complete


zmb3:
Thanks for the PR!

PR #6, opened 2016-01-19
Fix ExternalURL collections

brandur:
The current implementation seem to map these collections to a single struct
(the `ExternalURL` type) containing a Key/Value which in practice doesn't
unmarshal to anything. As dictated by the documentation [1], these are simple
objects mapping a string to a string and that's unlikely to change during this
API version, so I'd recommend using a simple map instead (for user
convenience).

The `ExternalURL` seems to be using the special JSON annotations of
`json:"{key}"`, which I would guess are supposed to help unmarshal an object,
but I couldn't find any record of them online (and as noted above, the singular
nature of the resources will prevent a successful unmarshaling anyway, I think
at the very least you'd want `[]ExternalURL`).

Note that `ExternalIDs` have the same problem, but are not addressed in this
pull.

[1] https://developer.spotify.com/web-api/get-playlist/

### DIFF
--- a/album.go
+++ b/album.go
@@ -49,7 +49,7 @@ type SimpleAlbum struct {
 	// widest first.
 	Images []Image `json:"images"`
 	// Known external URLs for this album.
-	ExternalURLs ExternalURL `json:"external_urls"`
+	ExternalURLs map[string]string `json:"external_urls"`
 }
 
 // Copyright contains the copyright statement associated with an album.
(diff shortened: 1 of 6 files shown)


zmb3:
You couldn't find anything on those "special JSON annotations" because there's no such thing.  I have no idea what I was thinking when I wrote that.

Thanks again for your contributions.  I'll go ahead and fix the `ExternalIDs` after merging this.

brandur:
> You couldn't find anything on those "special JSON annotations" because there's no such thing. I have no idea what I was thinking when I wrote that.

Hah, okay I'm glad that I wasn't going crazy ;) Thanks!

PR #11, opened 2016-07-28
Add support for recommendations based on seeds

sebgl:
Hi,

Thanks for this library ! Here is a contribution regarding the [Recommendations API](https://developer.spotify.com/web-api/get-recommendations/).

I tried to keep in line with the rest of the code. There are a few things I'd like to point out since they can be discussed :

* The big fat TrackAttributes struct is declared in its own file for better readability, but is only used by recommendations.go.
* Serializing these attributes to url values the usual way would represent a big function checking, for each of the 42 attributes, that it is not nil before inserting it into the url. In favor of code conciseness I opted for reflection instead, see the _setTrackAttributesValues_ function. There might be a small performance hit in doing reflection, that I did not evaluate. If you prefer the big if-based function I'm ok to change it.
* I'm not sure up to which point the package should validate user inputs, but it seemed right to check that _0 < numberOfSeeds <= 5_ as specified by Spotify.

I will be happy to make changes if needed, feel free to comment :)

### DIFF
--- a/recommendation.go
+++ b/recommendation.go
@@ -1,0 +1,130 @@
+// Copyright 2014, 2015 Zac Bergquist
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spotify
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+// Seeds contains IDs of artists, genres and/or tracks
+// to be used as seeds for recommendations
+type Seeds struct {
+	Artists []ID
+	Tracks  []ID
+	Genres  []string
+}
+
+// count returns the total number of seeds contained in s
+func (s Seeds) count() int {
+	return len(s.Artists) + len(s.Tracks) + len(s.Genres)
+}
+
+// Recommendations contains a list of recommended tracks based on seeds
+type Recommendations struct {
+	Seeds  []RecommendationSeed `json:"seeds"`
+	Tracks []SimpleTrack        `json:"tracks"`
+}
+
+// RecommendationSeed represents a recommendation seed after
+// being processed by the Spotify API
+type RecommendationSeed struct {
+	AfterFilteringSize int    `json:"afterFilteringSize"`
+	AfterRelinkingSize int    `json:"afterRelinkingSize"`
+	Endpoint           string `json:"href"`
+	ID                 ID     `json:"id"`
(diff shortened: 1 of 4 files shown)


zmb3:
This looks great, thanks!  I've been thinking for a while about how to make a good API around all of these different parameters.  I like the reflection approach and am not concerned at all about the performance hit - the HTTP call is going to dominate with or without reflection.

I would be interested in your opinions on setting track attributes though.  In my opinion, populating a pointer to an integral type is kind of a pain due to the need to create a temporary:

```
var ta TrackAttributes
tempo := 42
ta.TargetTempo = &tempo
``` 

I was toying with the idea of a map instead of a struct with pointer members:

```
attribs[spotify.TargetTempo] = 42
```

Or even something like a helper function that returns a `*float`.

Thoughts?

sebgl:
Yes, I also scratched my head on these structs of pointers.
Main problem is that we need a struct of optional fields, and the easy way to do that is to use pointers.

Maps would be a solution, but we have int and float values here (and string could come later on), meaning we should expose two maps?
```
intAttribs[spotify.TargetTempo] = 42
floatAttribs[spotify.Acousticness] = 0.12
```
Does not look very good.

I've been looking at the [go-github](https://github.com/google/go-github) library, which also contains many structs containing optional (pointers) types. What they did is declare helper functions like you suggested that [take a basic type and return its pointer](https://github.com/google/go-github/blob/master/github/github.go#L797-L807), to be used inline:
```
repo := &github.Repository{
    Name:    github.String("foo"),
    Private: github.Bool(true),
}
```
Looks better but I still find it confusing. As a user of the library I probably won't look around to check if these kind of helper functions exist in the library, but feed in my pointers directly.

Another approach could be to use setters only, and keep internal fields private:
```
var ta TrackAttributes
ta.SetTargetTempo(42)
ta.SetMinAcousticness(0.12)
```
This way we can still have control over int and float types.
Internally we can use two private maps instead of a struct of pointers, to avoid reflection when setting url values:
```
type TrackAttributes struct {
    intAttributes map[string]int
    floatAttributes map[string]float
}

func (ta TrackAttributes) SetTargetTempo(int tempo) {
    ta.intAttributes["targetTempo"] = tempo
}

func (ta TrackAttributes) GetTargetTempo() (int, error) {
    t, ok := ta.intAttributes["targetTempo"]
    if !ok {
        return -1, ErrAttributeNotSet   
    }
    return t
}
```
The disadvantage I see is that's a lot of "dumb" code to write, but overall I find it a much better solution. And the important point is the library user experience.

What do you think?

zmb3:
Setters with unexported fields sounds like the best idea.
How about a builder pattern?

```
ta := spotify.NewTrackAttributes()
   .TargetTempo(42)
   .Acousticness(0.12)
   .Energy(0.5)
```

sebgl:
Looks great !

Do you think getters are needed? I'm not sure this is necessary, but on the other hand it feels strange not to be able to.

It's too bad we can't do `tempo := ta.TargetTempo()` to get the value, when we have `ta.TargetTempo(42)` to set it.
Which leaves us with something like `tempo := ta.GetTargetTempo()` if we want to implement it.

zmb3:
If the attributes are only ever used as input, there's probably no need for getters as they are short lived, just for constructing the API call.

If some API calls return track attribute objects, then we should have getters so uses can get at the information in the response.

sebgl:
Commit modified to match what we said above.

PR #12, opened 2016-08-17
bugfix: does not get any audio features

fjukstad:
fdc4ddb does not get any audio features from spotify. Updated
GetAudioFeatures to actually retrieve audio features based on the song
IDs. If an ID is not found it will now throw an error not place
a nil value in the appropriate position. Since we're returning
a slice of AudioFeatures pointers I believe it is too cumbersome for the
caller to have to check these afterwards.

TODO: Update tests as well.

### DIFF
--- a/audio_features.go
+++ b/audio_features.go
@@ -105,25 +105,28 @@ const (
 
 // GetAudioFeatures queries the Spotify Web API for various
 // high-level acoustic attributes of audio tracks.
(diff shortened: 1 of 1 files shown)


zmb3:
Thanks for this, but I prefer keeping the nil element in the result so that result[n] always corresponds to id[n].

Merged #13 which fixes the bug.

fjukstad:
👍 great package by the way!

Issue #14, opened 2016-11-23
ReorderPlaylistTracks userID string


adrean:
It seems to me that `userID` should be a `string` in `ReorderPlaylistTracks`. It is expecting an `ID` instead

I'm learning go, and also new to the spotify API, so I might be missing something else :).
I'll gladly create the PR if you want

zmb3:
You're correct. Good catch. Please feel free to submit the PR.

PR #15, opened 2016-11-23
user id should be a string in ReorderPlaylistTracks

adrean:
issue #14


adrean:
you're right. There is already a test, the user id type was inferred because no variable used I guess. I changed it. Did you have something else in mind?

zmb3:
Thanks!

Issue #16, opened 2016-11-30
Follow type


cescoferraro:
This line is buggy
https://github.com/zmb3/spotify/blob/master/user.go#L210

ids parameter value is missing. and the type is mandatory.
https://developer.spotify.com/web-api/follow-artists-users/

this works
`spotifyURL := baseAddress + "me/following?type=artist&ids=" + strings.Join(toStringSlice(ids), ",")`

I can submit a PR for this, but want to know what do you consider being the best way of tackling this.

Both Follow and UNfollow use the same `modifyFollowers` and take  ids as varaidic arguments.
Shoud them take type as an argument? I think so. Following multiple people of different types would probably fail. SHould it be documented?

zmb3:
What if we make FollowArtists and FollowUsers, each of which would pass the
correct type to modifyFollowers, and then we can deprecate the original
follow funcs?

On Wed, Nov 30, 2016 at 10:06 AM cesco <notifications@github.com> wrote:

This line is buggy
https://github.com/zmb3/spotify/blob/master/user.go#L210

ids parameter value is missing. and the type is mandatory.
https://developer.spotify.com/web-api/follow-artists-users/

this works
spotifyURL := baseAddress + "me/following?type=artist&ids=" +
strings.Join(toStringSlice(ids), ",")

I can submit a PR for this, but want to know what do you consider being the
best way of tackling this.

Both Follow and UNfollow use the same modifyFollowers and take ids as
varaidic arguments.
Shoud them take type as an argument? I think so. Following multiple people
of different types would probably fail. SHould it be documented?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/zmb3/spotify/issues/16>, or mute the thread

cescoferraro:
i will write this later today @zmb3

zmb3:
Fixed by #18

PR #18, opened 2016-12-01
Feature/follow type

cescoferraro:
solves #16

### DIFF
--- a/user_test.go
+++ b/user_test.go
@@ -89,7 +89,7 @@ func TestFollowUsersMissingScope(t *testing.T) {
 	client := testClientString(http.StatusForbidden, json)
 	addDummyAuth(client)
 
-	err := client.Follow(ID("exampleuser01"))
+	err := client.FollowUser(ID("exampleuser01"))
 	if serr, ok := err.(Error); !ok {
 		t.Error("Expected insufficient client scope error")
 	} else {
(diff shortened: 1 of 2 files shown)


cescoferraro:
@zmb3 can you take a look at this?

user_test.go
```diff
@@ -89,7 +89,7 @@ func TestFollowUsersMissingScope(t *testing.T) {
 	client := testClientString(http.StatusForbidden, json)
 	addDummyAuth(client)
 
-	err := client.Follow(ID("exampleuser01"))
+	err := client.FollowUser(ID("exampleuser01"))
```
zmb3: In the tests we should verify that the `type` is set correctly.

There should be a `getLastRequest ()` that will give you the HTTP request that was made, and then you can inspect its URL values at `req.URL.Query()`.

user_test.go
```diff
@@ -109,7 +109,7 @@ func TestFollowUsersInvalidToken(t *testing.T) {
 	client := testClientString(http.StatusUnauthorized, json)
 	addDummyAuth(client)
 
-	err := client.Follow(ID("dummyID"))
+	err := client.FollowUser(ID("dummyID"))
```
zmb3: Could also add a test for `FollowArtist`

zmb3:
Thanks again, looks great.

Issue #20, opened 2017-01-13
Error when authenticating user


alexrs:
Running the auth example of the library returns the following error:

```BASH
2017/01/13 23:32:10 Post https://accounts.spotify.com/api/token: stream error: stream ID 1; REFUSED_STREAM
```

UPDATE:
If it is executed with `GODEBUG=http2client=0` it seems to work. This is the output I get when I set `GODEBUG=http2debug=2`:

```BASH
2017/01/13 23:33:57 http2: Transport failed to get client conn for accounts.spotify.com:443: http2: no cached connection was available
2017/01/13 23:33:57 http2: Transport creating client conn 0xc4200901a0 to 194.132.196.226:443
2017/01/13 23:33:57 http2: Framer 0xc4202f5800: wrote SETTINGS len=18, settings: ENABLE_PUSH=0, INITIAL_WINDOW_SIZE=4194304, MAX_HEADER_LIST_SIZE=10485760
2017/01/13 23:33:57 http2: Framer 0xc4202f5800: wrote WINDOW_UPDATE len=4 (conn) incr=1073741824
2017/01/13 23:33:57 http2: Transport encoding header ":authority" = "accounts.spotify.com"
2017/01/13 23:33:57 http2: Transport encoding header ":method" = "POST"
2017/01/13 23:33:57 http2: Transport encoding header ":path" = "/api/token"
2017/01/13 23:33:57 http2: Transport encoding header ":scheme" = "https"
2017/01/13 23:33:57 http2: Transport encoding header "content-type" = "application/x-www-form-urlencoded"
2017/01/13 23:33:57 http2: Transport encoding header "authorization" = "Basic NDEzZGY5N2FlOTZmNDliODgxN2Q1MzIyNzg4YjI0YzI6NDM5Mzk3Zjg1OGM4NGE3ZWJmOTUzMzFmZjA2NjA3ZDY="
2017/01/13 23:33:57 http2: Transport encoding header "content-length" = "379"
2017/01/13 23:33:57 http2: Transport encoding header "accept-encoding" = "gzip"
2017/01/13 23:33:57 http2: Transport encoding header "user-agent" = "Go-http-client/2.0"
2017/01/13 23:33:57 http2: Framer 0xc4202f5800: wrote HEADERS flags=END_HEADERS stream=1 len=158
[...]

PR #21, opened 2017-01-16
Change http client of oauth2 to force HTTP/1

Alexrs95:
As discussed in #20, oauth2 used HTTP/2 for authenticating the user. Now, the old `auth2.NoContext` has been substituted by `context.Context` and the default `http.Client` has been overrided as stated in the [documentation](https://golang.org/pkg/net/http/#pkg-overview) for using HTTP/1

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -68,15 +68,9 @@ const (
 //     client := a.NewClient(token)
 //
 type Authenticator struct {
-	config *oauth2.Config
+	config  *oauth2.Config
+	context context.Context
 }
-
-var (
-	tr = &http.Transport{
-		TLSNextProto: map[string]func(authority string, c *tls.Conn) http.RoundTripper{},
-	}
-	ctx = context.WithValue(context.TODO(), oauth2.HTTPClient, &http.Client{Transport: tr})
-)
 
 // NewAuthenticator creates an authenticator which is used to implement the
 // OAuth2 authorization flow.  The redirectURL must exactly match one of the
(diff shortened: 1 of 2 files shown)


Alexrs95:
Done! Now the context is created inside `NewAuthenticator` and stored inside `Authenticator`, and can be accessed using `authenticator.context`.

Should the user be able to pass the context as parameter to `NewAuthenticator`, and if it is `nil`, use this default context?

zmb3:
> Should the user be able to pass the context as parameter to NewAuthenticator, and if it is nil, use this default context?

Eventually, yes, the user should be able to pass a context into just about every exported func in the library.  This would be a breaking change though so we need to think about how to best do that.  Maybe use gopkg.in or do the work in a V2 branch.

I would suggest using `context.Background()` instead of `context.TODO()` here.

The other thing we need to do is make sure that `spotify.DefaultClient` is also setup to disable HTTP/2.  Would you mind adding this change?

Alexrs95:
I have changed `context.TODO()` by `context.Backgroudn()`.
Apparently, `spotify.DefaultClient` works as expected. I have tried the examples provided by the library and crating and adding songs to a playlist.

Do you want me to do any other change before sending the pull request?

zmb3:
> Apparently, `spotify.DefaultClient` works as expected. I have tried the examples provided by the library and crating and adding songs to a playlist.

Maybe they're running a newer version of nginx on these endpoints or something.  In any case, the DefaultClient could still try to use HTTP/2, which Spotify indicates is something they don't plan to support.  I'd rather have the default client behave the same as an authenticated one.

Alexrs95:
As far as I know, HTTP/2 is used when the server supports it, if not, HTTP/1.x is used. I don't know why when doing the authentication, it tries to do it using HTTP/2

zmb3:
Right, but by default the Go client is going to try HTTP/2 first, unless you set `TLSNextProto`.

Can you create an `init()` function and setup `TLSNextProto` on on the default client?

PR #23, opened 2017-01-24
Fix Invalid Type error in CurrentUsersFollowedArtistsOpt

wndhydrnt:
When passing something other than the default values to
`CurrentUsersFollowedArtistsOpt`, the client returns an error
like "Invalid type: artist?limit=50".


zmb3:
Thanks for catching this.  Can you add a test (or update an existing one) to cover this behavior?

wndhydrnt:
I'll see what I can do.

wndhydrnt:
@zmb3 I've added a test.

PR #25, opened 2017-02-07
added link to RapidAPI

phillipseymor:
Hey! I've added a link in your README to Spotify's functions page in RapidAPI. I've done this for a few Spotify SDKs to help those who are reading the READMEs, understand and test the API before use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,27 @@ Spotify
 This is a Go wrapper for working with Spotify's
 [Web API](https://developer.spotify.com/web-api/).
 
+<div style="margin: 25px;">
+<a href="https://rapidapi.com/package/SpotifyPublicAPI/functions?utm_source=SpotifyGitHub&utm_medium=button&utm_content=Vendor_GitHub" style="
+    all: initial;
+    background-color: #498FE1;
+    border-width: 0;
+    border-radius: 5px;
+    padding: 10px 20px;
+    color: white;
+    font-family: 'Helvetica';
+    font-size: 12pt;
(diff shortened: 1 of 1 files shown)


zmb3:
Would you mind pasting a screenshot here of what this looks like?

I have never heard of RapidAPI.  What does it provide that Spotify's Web API console doesn't?

phillipseymor:
Sure thing! Basically, RapidAPI is an abstraction layer for APIs. It allows devs to test out an API directly in our dashboard to ultimately understand what the response will look like. Most APIs don't have a robust web console like Spotify, so we want to make it easier for devs to connect to them. That being said, a user can also generate any API call in any language of their choice. We've had over 22,000 API calls with our Spotify package over the last few months.

screen shot 2017-02-07 at 2 34 12 pm

phillipseymor:
Essentially, before a dev implements your Go SDK, they'll be able to play around with what the response will look like. :)

zmb3:
Looks like a cool product, and I see the value for APIs that don't offer similar behavior, but I'm not sure my README is the right place to endorse it.

If I'm understanding correctly, the answer to my question _what does RapidAPI provide that Spotify's console doesnt?_ is unfortunately _not much_.

I'm also not a huge fan of raw HTML in markdown - I think it makes it hard to read in text editors where developers spend most of there time.

I would be open to adding a smaller section to the bottom similar to:

> You may find tools such as Spotify's Web API Console [link] or Rapid API [link] valuable for experimenting with the API.

zmb3:
Thanks!

PR #26, opened 2017-03-03
update Token on NewClient

jgsqware:
On NewClient, in case of ExpiredToken, regenerate a new Token and update the current Token

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -145,9 +145,14 @@ func (a Authenticator) Exchange(code string) (*oauth2.Token, error) {
 }
 
 // NewClient creates a Client that will use the specified access token for its API requests.
-func (a Authenticator) NewClient(token *oauth2.Token) Client {
+func (a Authenticator) NewClient(token *oauth2.Token) (Client, error) {
 	client := a.config.Client(a.context, token)
+	t, err := client.Transport.(*oauth2.Transport).Source.Token()
+	*token = *t
+	if err != nil {
+		return Client{}, err
+	}
 	return Client{
 		http: client,
-	}
+	}, nil
 }
--- a/auth.go
+++ b/auth.go
@@ -145,14 +145,19 @@ func (a Authenticator) Exchange(code string) (*oauth2.Token, error) {
 }
 
 // NewClient creates a Client that will use the specified access token for its API requests.
-func (a Authenticator) NewClient(token *oauth2.Token) (Client, error) {
+func (a Authenticator) NewClient(token *oauth2.Token) Client {
 	client := a.config.Client(a.context, token)
-	t, err := client.Transport.(*oauth2.Transport).Source.Token()
-	*token = *t
-	if err != nil {
-		return Client{}, err
-	}
 	return Client{
 		http: client,
-	}, nil
+	}
+}
+
+// Token return the current client token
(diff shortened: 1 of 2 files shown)


jgsqware:
AppVeyor apparently has not been relauched after the last commit. The error AppVeyor show is fixed

zmb3:
I'm not sure that this is necessary.  The client we obtain from the oauth2 package will refresh the token automatically.

Per the[ docs](https://godoc.org/golang.org/x/oauth2#Config.Client):
> Client returns an HTTP client using the provided token. The token will auto-refresh as necessary.

jgsqware:
Yep, but the retrieve client don't have the token accessible. I use it to
save to token for further request

Le ven. 3 mars 2017 16:01, Zac Bergquist <notifications@github.com> a
écrit :

zmb3:
I might just not understand your use case.  You shouldn't need to access the token as long as you use the client returned by `NewClient` - everything is handled transparently to you.

What do you need access to the token for?  We might be able to make it easier on you.

jgsqware:
I use the lib in a cli.
The first authorization save the token for later use, then on next run a
reuse that token and NewClient renew it. But I cannot save that renewed
token.

Le ven. 3 mars 2017 16:13, Zac Bergquist <notifications@github.com> a
écrit :

zmb3:
Okay, that makes sense.

What if we added a `func (c *Client) Token() *oauth2.Token` instead?  I feel like thats a more explicit API than modifying the token passed in to `NewClient`.

jgsqware:
Ok I can update my PR if needed

Le ven. 3 mars 2017 16:18, Zac Bergquist <notifications@github.com> a
écrit :

examples/authenticate/authenticate.go
```diff
@@ -60,6 +60,9 @@ func completeAuth(w http.ResponseWriter, r *http.Request) {
 	}
 	// use the token to get an authenticated client
 	client := auth.NewClient(tok)
+	if err != nil {
```
zmb3: This needs to be removed.

auth.go
```diff
@@ -151,3 +151,13 @@ func (a Authenticator) NewClient(token *oauth2.Token) Client {
 		http: client,
 	}
 }
+
+// Token return the current client token
+func (c *Client) Token() (*oauth2.Token, error) {
+	t, err := c.http.Transport.(*oauth2.Transport).Source.Token()
```
zmb3: Let's check the type assertion just to be safe:
```
transport, ok := c.http.Transport.(*ouath2.Transport)
if !ok {
   return nil, errors.New("...")
}
t, err := transport.Source.Token()
// ...
```

auth.go
```diff
@@ -151,3 +151,13 @@ func (a Authenticator) NewClient(token *oauth2.Token) Client {
 		http: client,
 	}
 }
+
+// Token return the current client token
```
zmb3: How about: 
`// Token gets the client's current token.`

chances:
I have a similar use case: Storing the refresh token in a database for future use.

Anything we can do to help get this PR merged?

zmb3:
Sorry this slipped through the cracks. I never saw the review comments addressed. Looks good!

PR #29, opened 2017-04-04
Initial work to add Web API Connect Endpoint

freman:
See: https://developer.spotify.com/web-api/web-api-connect-endpoint-reference/

Still needs some tests, more than willing to take feedback on naming or whatever.

### DIFF
--- a/player.go
+++ b/player.go
@@ -1,0 +1,503 @@
+package spotify
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/url"
+	"strconv"
+)
+
+var (
+	// ErrTemporarilyUnavailable When the device is temporarily unavailable the request will
+	// return this error the client should retry the request after 5 seconds,
+	// but no more than at most 5 retries.
+	ErrTemporarilyUnavailable = errors.New("Device is temporarily unavailable")
+
+	// ErrDeviceNotFound The requested device is not found
+	ErrDeviceNotFound = errors.New("Device is not found")
+
+	// ErrNotPremium The user making the request is non-premium
+	ErrNotPremium = errors.New("User is not premium")
+)
+
+// PlayerDevice contains information about a device that a user can play music on
+type PlayerDevice struct {
+	// ID of the device. This may be empty.
+	ID string `json:"id"`
+	// IsActive If this device is the currently active device.
+	IsActive bool `json:"is_active"`
+	// IsRestricted Whether controlling this device is restricted. At present if
+	// this is "true" then no Web API commands will be accepted by this device.
+	IsRestricted bool `json:"is_restricted"`
+	// Name The name of the device.
+	Name string `json:"name"`
+	// Type Device type, such as "Computer", "Smartphone" or "Speaker".
+	Type string `json:"type"`
+	// Volume The current volume in percent.
+	Volume int `json:"volume_percent"`
+}
+
+// PlayerDeviceList is a list of devices
+type PlayerDeviceList struct {
+	PlayerDevices []PlayerDevice `json:"devices"`
+}
+
+// PlayerState contains information about the current playback.
+type PlayerState struct {
+	CurrentlyPlaying
+	// Device The device that is currently active
+	Device string `json:"device"`
+	// ShuffleState Shuffle is on or off
+	ShuffleState bool `json:"shuffle_state"`
+	// RepeatState off, track, context
+	RepeatState string `json:"repeat_state"`
+}
+
+// Context is the playback context
+type Context struct {
+	// ExternalURLs of the context, or null if not available.
+	ExternalURLs map[string]string `json:"external_urls"`
+	// Endpoint of the context, or null if not available.
+	Endpoint string `json:"href"`
+	// Type of the item's context. Can be one of album, artist or playlist.
+	Type string `json:"type"`
+	// URI of the context.
+	URI URI `json:"uri"`
+}
+
+// CurrentlyPlaying contains the information about currently playing items
+type CurrentlyPlaying struct {
+	// Timestamp when data was fetched
+	Timestamp int `json:"timestamp"`
+	// Context current context
+	Context Context `json:"context"`
+	// Progress into the currently playing track.
+	Progress string `json:"progress_ms"`
+	// IsPlaying If something is currently playing.
+	IsPlaying bool `json:"is_playing"`
+	// The currently playing track. Can be null.
+	Item *FullTrack `json:"Item"`
+}
+
+type Offset struct {
+	//Position is zero based and can’t be negative.
+	Position int `json:"position,omitempty"`
+	// URI is a string representing the uri of the item to start at.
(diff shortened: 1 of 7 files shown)


freman:
Not sure how to fix AppVeyer, it works in my checkout when I run it in the freman namespace (And when I rename the path to zmb3 namespace)

player.go
```diff
@@ -0,0 +1,493 @@
...
+	"net/url"
+	"strconv"
+)
+
+// PlayerDevice contains information about a device that a user can play music on
+type PlayerDevice struct {
+	// ID of the device. This may be empty.
+	ID string `json:"id"`
+	// IsActive If this device is the currently active device.
+	IsActive bool `json:"is_active"`
```
zmb3: I would probably just call these `Active` and `Restricted`.

player.go
```diff
@@ -0,0 +1,493 @@
...
+	// IsRestricted Whether controlling this device is restricted. At present if
+	// this is "true" then no Web API commands will be accepted by this device.
+	IsRestricted bool `json:"is_restricted"`
+	// Name The name of the device.
+	Name string `json:"name"`
+	// Type Device type, such as "Computer", "Smartphone" or "Speaker".
```
zmb3: Try to make the comments read more like sentences:

> Type of the device, such as "Computer", "Smartphone",  or "Speaker"

player.go
```diff
@@ -0,0 +1,493 @@
...
+	Type string `json:"type"`
+	// Volume The current volume in percent.
+	Volume int `json:"volume_percent"`
+}
+
+// PlayerDeviceList is a list of devices
+type PlayerDeviceList struct {
```
zmb3: If you're not attaching any methods to `PlayerDeviceList`, I would prefer just to use `[]PlayerDevice` everywhere.  This way it's immediately obvious that we're returning a list (slice) of devices.

player.go
```diff
@@ -0,0 +1,493 @@
...
+	// Device The device that is currently active
+	Device PlayerDevice `json:"device"`
+	// ShuffleState Shuffle is on or off
+	ShuffleState bool `json:"shuffle_state"`
+	// RepeatState off, track, context
+	RepeatState string `json:"repeat_state"`
+}
+
+// Context is the playback context
+type Context struct {
```
zmb3: Maybe we should call this `PlaybackContext` so that it doesn't seem related to `context.Context`

player.go
```diff
@@ -0,0 +1,493 @@
...
+	// Progress into the currently playing track.
+	Progress int `json:"progress_ms"`
+	// IsPlaying If something is currently playing.
+	IsPlaying bool `json:"is_playing"`
+	// The currently playing track. Can be null.
+	Item *FullTrack `json:"Item"`
+}
+
+type Offset struct {
+	//Position is zero based and can’t be negative.
```
zmb3: Space after `//`

What are the units on position?

player.go
```diff
@@ -0,0 +1,493 @@
...
+	Context Context `json:"context"`
+	// Progress into the currently playing track.
+	Progress int `json:"progress_ms"`
+	// IsPlaying If something is currently playing.
+	IsPlaying bool `json:"is_playing"`
+	// The currently playing track. Can be null.
+	Item *FullTrack `json:"Item"`
+}
+
+type Offset struct {
```
zmb3: `PlaybackOffset` ?

player.go
```diff
@@ -0,0 +1,493 @@
...
+	//Position is zero based and can’t be negative.
+	Position int `json:"position,omitempty"`
+	// URI is a string representing the uri of the item to start at.
+	URI URI `json:"uri,omitempty"`
+}
+
+type PlayOptions struct {
+	// DeviceID The id of the device this command is targeting. If not
+	// supplied, the user's currently active device is the target.
+	DeviceID *string `json:"-"`
```
zmb3: Why `json:"-"`?
Is this in the form of a Spotify ID?  If so, we have a type for that.

player.go
```diff
@@ -0,0 +1,493 @@
...
+// TransferPlayback transfers playback to a new device and determine if
+// it should start playing. This call requires authorization.
+//
+// Note that a value of false for the play parameter when also transferring
+// to another device_id will not pause playback. To ensure that playback is
+// paused on the new device you should send a pause command to the currently
+// active device before transferring to the new device_id.
+//
+// Requires the ScopeUserModifyPlaybackState in order to modify the player state
+func (c *Client) TransferPlayback(device_ids []string, play bool) error {
```
zmb3: Don't use underscores, call the argument `deviceIDs`.

Again, are these strings or Spotify IDs?

player.go
```diff
@@ -0,0 +1,493 @@
...
+
+// Previous skips to the Previous track in the user's queue in the user's
+// currently active device. This call requires authorization.
+//
+// Requires the ScopeUserModifyPlaybackState in order to modify the player state
+func (c *Client) Previous() error {
+	return c.NextOpt(nil)
+}
+
+// PreviousOpt Like Previous but with more options
```
zmb3: PreviousOpt _is_ like Previous but with more options.

player.go
```diff
@@ -0,0 +1,493 @@
...
+//
+// Requires the ScopeUserModifyPlaybackState in order to modify the player state
+func (c *Client) Repeat(state string) error {
+	return c.RepeatOpt(state, nil)
+}
+
+// RepeatOpt Like Repeat but with more options
+//
+// Only expects PlayOptions.DeviceID, all other options will be ignored
+func (c *Client) RepeatOpt(state string, opt *PlayOptions) error {
```
zmb3: What is `state`?
Should we remove `*PlayOptions` and accept an optional device ID instead?

player.go
freman: Ok

player.go
freman: Ok

player.go
freman: Was more for the json.Unmarshal, but Ok

player.go
freman: Ok

player.go
freman: Ok

player.go
freman: Surprised the linter doesn't pick that up.

RE Units: The api doesn't specify, in the context of their example it's the offset of a track in a playlist or list of uris, but I suspect it's also milliseconds when given a track.

player.go
freman: Stops json.Marshal from marshalling the DeviceID which all requests expect in a get param not as part of the post/put - Being used for the convenience of it being optional in all the functions (so far only really one) that actually take these options vs using a seperate positional argument.

player.go
freman: Yeh sorry, was tired

player.go
freman: fixed

player.go
freman: I was going for consistency (Ie: things suffixed 'opt' take opts...) but if you like I can do that.

PR #31, opened 2017-04-08
Add support for Get Audio Analysis

therealpenguin:
AudioAnalysis contains a detailed audio analysis for a single track identified by its unique Spotify ID. See https://developer.spotify.com/web-api/get-audio-analysis/ 

Spotify's documentation is currently missing the object model for the AudioAnalysis. See https://github.com/spotify/web-api/issues/317

Also see The Echo Nest documentation https://web.archive.org/web/20160528174915/http://developer.echonest.com/docs/v4/_static/AnalyzeDocumentation.pdf

### DIFF
--- a/audio_analysis.go
+++ b/audio_analysis.go
@@ -8,29 +8,24 @@ import (
 
 // AudioAnalysis contains a detailed audio analysis for a single track identified by its unique Spotify ID.
 // See https://developer.spotify.com/web-api/get-audio-analysis/
-//
-// Spotify's documentation is currently missing the object model for the AudioAnalysis.
-// See https://github.com/spotify/web-api/issues/317
-//
-// Also see The Echo Nest documentation
-// https://web.archive.org/web/20160528174915/http://developer.echonest.com/docs/v4/_static/AnalyzeDocumentation.pdf
 type AudioAnalysis struct {
-	Bars     []Measure          `json:"bars"`
-	Beats    []Measure          `json:"beats"`
+	Bars     []Marker           `json:"bars"`
+	Beats    []Marker           `json:"beats"`
 	Meta     AudioAnalysisMeta  `json:"meta"`
 	Sections []Section          `json:"sections"`
 	Segments []Segment          `json:"segments"`
-	Tatums   []Measure          `json:"tatums"`
+	Tatums   []Marker           `json:"tatums"`
 	Track    AudioAnalysisTrack `json:"track"`
 }
 
-// Measure represents beats, bars, tatums and are used in segments and sections descriptions.
-type Measure struct {
+// Marker represents beats, bars, tatums and are used in segment and section descriptions.
+type Marker struct {
 	Start      float64 `json:"start"`
 	Duration   float64 `json:"duration"`
 	Confidence float64 `json:"confidence"`
 }
 
+// AudioAnalysisMeta describes details about Spotify's analysis of the track
 type AudioAnalysisMeta struct {
 	AnalyzerVersion string  `json:"analyzer_version"`
 	Platform        string  `json:"platform"`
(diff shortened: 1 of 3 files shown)


therealpenguin:
Should I include (copy and paste) some of the documentation from The Echo Nest pdf?

audio_analysis.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// AudioAnalysis contains a detailed audio analysis for a single track identified by its unique Spotify ID.
+// See https://developer.spotify.com/web-api/get-audio-analysis/
+//
+// Spotify's documentation is currently missing the object model for the AudioAnalysis.
```
zmb3: I would omit this.  Eventually they'll add documentation and we'll more than likely forget to come back and remove this.

audio_analysis.go
```diff
@@ -0,0 +1,119 @@
...
+	EchoprintString         string  `json:"echoprintstring"`
+	EchoprintVersion        float64 `json:"echoprint_version"`
+	SynchString             string  `json:"synchstring"`
+	SynchVersion            float64 `json:"synch_version"`
+	RhythmString            string  `json:"rhythmstring"`
+	RhythmVersion           float64 `json:"rhythm_version"`
+}
+
+// GetAudioAnalysis queries the Spotify web API for an audio analysis of a single track
+// If an object is not found, a nil value is returned in the appropriate position.
```
zmb3: This comment doesn't seem correct.  It only returns a single value, so the _appropriate position_ part is not relevant.

audio_analysis_test.go
```diff
@@ -0,0 +1,142 @@
+package spotify
+
+import (
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+const FIELDS_DIFFER_TEMPLATE = "Actual response is not the same as expected response on field %s"
```
zmb3: The `ALL_CAPS` naming convention is not idiomatic Go.  Additionally, this constant need not be exported.

Rename this to `fieldsDifferTemplate`.

audio_analysis.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Measure represents beats, bars, tatums and are used in segments and sections descriptions.
+type Measure struct {
+	Start      float64 `json:"start"`
+	Duration   float64 `json:"duration"`
+	Confidence float64 `json:"confidence"`
+}
+
+type AudioAnalysisMeta struct {
```
zmb3: Can you add doc comments for `AudioAnalysisMeta`, `Section`, `Segment`, and `AudioAnalysisTrack`?

therealpenguin:
Thank you for your helpful review 👍 

Moving the analysis to it's own package would help with the things you commented on - it is a bit weird to type AudioAnalysisTrack.

Am I right in thinking it would require a small amount of refactoring in the existing package? Things like exporting the HTTP client of `spotify.Client` and perhaps the baseAddress.

audio_analysis.go
```diff
@@ -41,21 +36,24 @@ type AudioAnalysisMeta struct {
 	InputProcess    string  `json:"input_process"`
 }
 
+// A Section represents a large variation in rhythm or timbre, e.g. chorus, verse, bridge, guitar solo, etc.
```
zmb3: `go lint` should warn you about this, but these doc comments should start with the name of the type they are documenting.

For example, this could say:

> Section represents a large variation...

Sorry if it seems I'm being overly pedantic, but I'd prefer not to introduce lint warnings if we can avoid it :-)

zmb3:
> Am I right in thinking it would require a small amount of refactoring in the existing package? Things like exporting the HTTP client of spotify.Client and perhaps the baseAddress.

Yes, but I'd prefer not to export the HTTP client, so my suggestion would be to move the analysis _types_ into a subpackage, but leave the main logic (which appears to be a single function) in `package spotify`.

So you'd end up with something like this:

```
func (c *Client) GetAudioAnalysis(id ID) (*analysis.Info, error) 
```

I think `analysis.Info` or `analysis.Result` would be a good name (`analysis.
Analysis` stutters a bit)

theckman:
@zmb3 I just took a look at hammering this out to help @therealpenguin since I think it's pretty cool and might use it, but it's a bit of a Yak Shave. I agree that `analysis.Foo` would be way more awesome, but it seems that some of the types for Audio Analysis have dependencies on types within the main package (`Key` and `Mode`):

* https://github.com/zmb3/spotify/pull/31/files#diff-e59e588f083230f3cdf157fa07a7b53fR81
* https://github.com/zmb3/spotify/pull/31/files#diff-e59e588f083230f3cdf157fa07a7b53fR83

I'm not sure what our preferred path may be, but noodling on it I think I've come with three potential solutions:

1. do nothing major; enhance comments based on outstanding feedback and maybe improve type names
2. move things to the suggested `analysis` package, and duplicate the `Key` and `Mode` types there
3. break API and migrate the `Key` and `Mode` to the `analysis` package

Option 1 or 2 seem the most reasonable.

zmb3:
Option 1 sounds good to me.

the separate package was just a suggestion without thinking too hard about
the details. Thanks for helping out!

audio_analysis.go
theckman: @zmb3 This actually appears to be valid:

```
$ golint .
audio_features.go:81:2: exported const C should have comment (or a comment on this block) or be unexported
audio_features.go:104:2: exported const Minor should have comment (or a comment on this block) or be unexported
player.go:75:6: exported type PlayOptions should have comment or be unexported
player.go:160:1: comment on exported method Client.PlayerCurrentlyPlayingOpt should be of the form "PlayerCurrentlyPlayingOpt ..."
```

theckman:
@zmb3 I opened another PR with @therealpenguin's commits included. It was rebased against master, so it should absolutely merge cleanly (not that there were any conflicts).

zmb3:
Closed via #36

PR #39, opened 2017-05-21
Added automatic-retries

dsoprea:
The Spotify API imposes rate-limits but this is obscured by the client. I added the ability to enable automatic-retries by the client (this has to be specifically enabled though). Before I did this, I had to refactor all of the requests to use common request mechanisms. These common calls:

1. Use an alternative implementation of `decodeError` that extracts the retry-interval from the headers and sets it into the `Client` struct.
2. Determine if we're to automatically retry, waits the provided amount, and then tries again.

I also added a small thing that sets our error-message to the description for the HTTP error-code if the request failed but the error-message is empty. This was really confusing when I tried to add a bunch of tracks to a playlist and it failed for no clear reason.

### DIFF
--- a/library.go
+++ b/library.go
@@ -54,13 +54,9 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
 	if err != nil {
 		return err
 	}
-	resp, err := c.http.Do(req)
+	err = c.execute(req)
 	if err != nil {
 		return err
-	}
-	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		return decodeError(c, resp)
 	}
 	return nil
 }
--- a/library.go
+++ b/library.go
@@ -14,10 +14,10 @@ func (c *Client) UserHasTracks(ids ...ID) ([]bool, error) {
 		return nil, errors.New("spotify: UserHasTracks supports 1 to 50 IDs per call")
 	}
 	spotifyURL := fmt.Sprintf("%sme/tracks/contains?ids=%s", baseAddress, strings.Join(toStringSlice(ids), ","))
-	
+
 	var result []bool
 
-	err := c.Get(spotifyURL, &result)
+	err := c.get(spotifyURL, &result)
 	if err != nil {
 		return nil, err
 	}
--- a/library.go
+++ b/library.go
@@ -1,7 +1,6 @@
 package spotify
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -15,16 +14,14 @@ func (c *Client) UserHasTracks(ids ...ID) ([]bool, error) {
 		return nil, errors.New("spotify: UserHasTracks supports 1 to 50 IDs per call")
 	}
 	spotifyURL := fmt.Sprintf("%sme/tracks/contains?ids=%s", baseAddress, strings.Join(toStringSlice(ids), ","))
-	resp, err := c.http.Get(spotifyURL)
+	
+	var result []bool
+
+	err := c.Get(spotifyURL, &result)
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		return nil, decodeError(resp.Body)
-	}
-	var result []bool
-	err = json.NewDecoder(resp.Body).Decode(&result)
+
 	return result, err
 }
 
@@ -63,7 +60,7 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return decodeError(resp.Body)
+		return decodeError(c, resp)
 	}
 	return nil
 }
--- a/page.go
+++ b/page.go
@@ -88,7 +88,7 @@ type CategoryPage struct {
 
 // getPage GETs the data at the specified URL and unmarshals it into page.
 func (c *Client) getPage(url string, page interface{}) error {
-	err := c.Get(url, page)
+	err := c.get(url, page)
 	if err != nil {
 		return err
 	}
--- a/page.go
+++ b/page.go
@@ -1,7 +1,6 @@
 package spotify
 
 import (
-	"encoding/json"
 	"errors"
 )
 
@@ -89,10 +88,10 @@ type CategoryPage struct {
 
 // getPage GETs the data at the specified URL and unmarshals it into page.
 func (c *Client) getPage(url string, page interface{}) error {
-	resp, err := c.http.Get(url)
+	err := c.Get(url, page)
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
-	return json.NewDecoder(resp.Body).Decode(page)
+
+	return nil
 }
--- a/page.go
+++ b/page.go
@@ -85,13 +85,3 @@ type CategoryPage struct {
 	basePage
 	Categories []Category `json:"items"`
 }
-
-// getPage GETs the data at the specified URL and unmarshals it into page.
-func (c *Client) getPage(url string, page interface{}) error {
-	err := c.get(url, page)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
--- a/spotify.go
+++ b/spotify.go
@@ -125,7 +125,7 @@ func (c *Client) decodeError(resp *http.Response) error {
 	}
 
 	if len(responseBody) == 0 {
-		return fmt.Errorf("spotify: there was an HTTP error (%d '%s') but the body was empty", resp.StatusCode, http.StatusText(resp.StatusCode))
+		return fmt.Errorf("spotify: HTTP %d: %s (body empty)", resp.StatusCode, http.StatusText(resp.StatusCode))
 	}
 
 	buf := bytes.NewBuffer(responseBody)
@@ -143,7 +143,7 @@ func (c *Client) decodeError(resp *http.Response) error {
 		if retrySecondsRaw != "" {
 			retrySeconds, err := strconv.ParseInt(retrySecondsRaw, 10, 32)
 			if err != nil {
-				return fmt.Errorf("could not parse retry seconds: %s", retrySecondsRaw)
+				return fmt.Errorf("spotify: could not parse retry seconds: %s", retrySecondsRaw)
 			}
 
 			if retrySeconds == 0 {
@@ -159,7 +159,7 @@ func (c *Client) decodeError(resp *http.Response) error {
 		// some of the arguments directly in the HTTP query and the URL ends-up
 		// being too long.
 
-		e.E.Message = fmt.Sprintf("spotify: there was an HTTP error (%d '%s') but we received an empty error", resp.StatusCode, http.StatusText(resp.StatusCode))
+		e.E.Message = fmt.Sprintf("spotify: unexpected HTTP %d: %s (empty error)", resp.StatusCode, http.StatusText(resp.StatusCode))
 	}
 
 	return e.E
--- a/spotify.go
+++ b/spotify.go
@@ -180,7 +180,7 @@ func (c *Client) SetAutoRetry(flag bool) {
 	c.autoRetry = flag
 }
 
-func (c *Client) ExecuteOpt(req *http.Request, needsStatus int, result interface{}) error {
+func (c *Client) executeOpt(req *http.Request, needsStatus int, result interface{}) error {
 	for {
 		resp, err := c.http.Do(req)
 		if err != nil {
@@ -209,11 +209,11 @@ func (c *Client) ExecuteOpt(req *http.Request, needsStatus int, result interface
 	return nil
 }
 
-func (c *Client) Execute(req *http.Request) error {
-	return c.ExecuteOpt(req, 0, nil)
+func (c *Client) execute(req *http.Request) error {
+	return c.executeOpt(req, 0, nil)
 }
 
-func (c *Client) Get(url string, result interface{}) error {
+func (c *Client) get(url string, result interface{}) error {
 	for {
 		resp, err := c.http.Get(url)
 		if err != nil {
--- a/spotify.go
+++ b/spotify.go
@@ -118,7 +118,7 @@ func (e Error) Error() string {
 }
 
 // decodeError decodes an Error from an io.Reader.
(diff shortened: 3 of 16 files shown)


spotify.go
```diff
@@ -55,6 +68,10 @@ func init() {
 	DefaultClient.http.Transport = tr
 }
 
+func SetAutoRetry(flag bool) {
```
zmb3: This should be a property of `Client` and not global to the package.

spotify.go
```diff
@@ -107,14 +124,48 @@ func (e Error) Error() string {
...
+
 	var e struct {
 		E Error `json:"error"`
 	}
-	err := json.NewDecoder(r).Decode(&e)
+	err = json.NewDecoder(buf).Decode(&e)
 	if err != nil {
-		return errors.New("spotify: couldn't decode error")
+		if len(responseBody) == 0 {
+			errors.New("spotify: could not decode empty body")
```
zmb3: - You're not returning an error here, simply creating one and throwing it away.
- It seems like you could check for an empty response body before trying to decode JSON.
- Since you're already checking for an empty body ahead of time, there's no need to put `len(responseBody)` in the error message
- Let's try to return the HTTP status code in the error if we can't decode it

spotify.go
```diff
@@ -107,14 +124,48 @@ func (e Error) Error() string {
...
+		} else {
+			return fmt.Errorf("spotify: couldn't decode error: (%d) [%s]", len(responseBody), responseBody)
+		}
 	}
+
+	if e.E.Error() == rateLimitExceededErrorMessage {
```
zmb3: I'm not crazy about comparing the error string as this could change at any time and break the library.  It looks like Spotify returns a 429 if the rate limits are exceeded - can you check this instead?

spotify.go
```diff
@@ -107,14 +124,48 @@ func (e Error) Error() string {
...
+		retrySecondsRaw := resp.Header.Get("Retry-After")
+		if retrySecondsRaw != "" {
+			retrySeconds, err := strconv.ParseInt(retrySecondsRaw, 10, 32)
+			if err != nil {
+				return fmt.Errorf("could not parse retry seconds: %s", retrySecondsRaw)
+			} else if retrySeconds == 0 {
```
zmb3: If an `if` statement ends with a `return`, omit the `else`:

```
if err != nil {
   return fmt.Errorf(...)
}
if retrySeconds == 0 {
   ...
}
```

spotify.go
```diff
@@ -124,6 +175,74 @@ func decodeError(r io.Reader) error {
...
+		if err != nil {
+			return err
+		}
+
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK || needsStatus != 0 && resp.StatusCode != needsStatus {
+			errorMessage := decodeError(c, resp)
+
+			if errorMessage.Error() == rateLimitExceededErrorMessage && autoRetry {
```
zmb3: See above re: comparing the error message string.

spotify.go
```diff
@@ -124,6 +175,74 @@ func decodeError(r io.Reader) error {
 // authenticate, you can use `DefaultClient`.
 type Client struct {
 	http *http.Client
+	retryDuration time.Duration
+}
+
+func (c *Client) ExecuteOpt(req *http.Request, needsStatus int, result interface{}) (err error) {
```
zmb3: - Does this need to be exported?
- What does `needsStatus` do?
- It seems like we should have a maximum number of retries to prevent getting stuck in an infinite loop.

spotify.go
```diff
@@ -124,6 +175,74 @@ func decodeError(r io.Reader) error {
...
+			}
+		}
+
+		break
+	}
+
+	return nil
+}
+
+func (c *Client) Execute(req *http.Request) (err error) {
```
zmb3: - Remove the named `err` return value.
- Can we unexport this?

spotify.go
```diff
@@ -124,6 +175,74 @@ func decodeError(r io.Reader) error {
...
+func (c *Client) ExecuteOpt(req *http.Request, needsStatus int, result interface{}) (err error) {
+	for {
+		resp, err := c.http.Do(req)
+		if err != nil {
+			return err
+		}
+
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK || needsStatus != 0 && resp.StatusCode != needsStatus {
```
zmb3: Are we certain that all endpoints return `http.StatusOK` on success?  I think some of the `PUT`/`DELETE`s will return a 204, which we should treat as a success.

spotify.go
```diff
@@ -161,25 +280,23 @@ func (c *Client) NewReleasesOpt(opt *Options) (albums *SimpleAlbumPage, err erro
 			spotifyURL += "?" + params
 		}
 	}
+
+// TODO(dustin): Doesn't currently support retrying because this is more complicate than all of the other Get() references that we've already replaced with our standard calls. This would require us to duplicate the functionality out of the Get() function. However, we suspect that this can be simplified, though we'll need a second opinion before we make any changes.
```
zmb3: This can definitely be simplified, but let's come back to it on the 2nd pass.

spotify.go
dsoprea: Done.

spotify.go
dsoprea: `needsStatus` is a secondary success code (which can be set to 0 to be ignored). In most cases we only check whether we got OK. In several of the non-GET requests (often for create calls) we also checked that it returned "created":

```
github.com/zmb3/spotify$ grep ExecuteOpt -R .
./spotify.go:func (c *Client) ExecuteOpt(req *http.Request, needsStatus int, result interface{}) (err error) {
./spotify.go:	return c.ExecuteOpt(req, 0, nil)
./playlist.go:	err = c.ExecuteOpt(req, http.StatusCreated, &p)
./playlist.go:	err = c.ExecuteOpt(req, http.StatusCreated, &body)
./playlist.go:	err = c.ExecuteOpt(req, 0, &result)
./playlist.go:	err = c.ExecuteOpt(req, 0, &result)
```

I'm not sure. If the server was telling us to wait and we threw an error or did something other than wait then I think we'd be interfering. It'd be inappropriate to apply our own interpretation of what we `think` we should do.

spotify.go
dsoprea: See above comment. `needsStatus` is provided as 204/`http.StatusCreated` in the couple of cases it is expected.

spotify.go
dsoprea: Done.

spotify.go
dsoprea: Done, though named return-values are used in the `NewReleases` methods.

I'll update these calls to not be exported in the next few minutes.

spotify.go
dsoprea: If that's what you prefer. Done.

spotify.go
dsoprea: Done.

spotify.go
dsoprea: Done.

dsoprea:
I just updated to not export the request functions.

The interval returned in the header is in seconds. In my experience, it's usually just three ("3"). I add (1) to this to account for rounding error, just in case. 

There's a bug in the server where the returned interval might often be (0). In this case I use (5), which is configured as a constant.

library.go
```diff
@@ -63,7 +60,7 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return decodeError(resp.Body)
+		return decodeError(c, resp)
```
zmb3: Let's make `decodeError` a method on `Client`.

page.go
```diff
@@ -89,10 +88,10 @@ type CategoryPage struct {
 
 // getPage GETs the data at the specified URL and unmarshals it into page.
 func (c *Client) getPage(url string, page interface{}) error {
```
zmb3: Can we remove `getPage` now since it's functionally equivalent to the `get` method you've added?

spotify.go
```diff
@@ -107,14 +118,50 @@ func (e Error) Error() string {
...
-		return errors.New("spotify: couldn't decode error")
+		return fmt.Errorf("spotify: couldn't decode error: (%d) [%s]", len(responseBody), responseBody)
+	}
+
+	if resp.StatusCode == rateLimitExceededStatusCode {
+		retrySecondsRaw := resp.Header.Get("Retry-After")
+		if retrySecondsRaw != "" {
+			retrySeconds, err := strconv.ParseInt(retrySecondsRaw, 10, 32)
+			if err != nil {
+				return fmt.Errorf("could not parse retry seconds: %s", retrySecondsRaw)
```
zmb3: Prefix this error with `spotify:`

spotify.go
```diff
@@ -107,14 +118,50 @@ func (e Error) Error() string {
...
 // decodeError decodes an Error from an io.Reader.
-func decodeError(r io.Reader) error {
+func decodeError(c *Client, resp *http.Response) error {
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if len(responseBody) == 0 {
+		return fmt.Errorf("spotify: there was an HTTP error (%d '%s') but the body was empty", resp.StatusCode, http.StatusText(resp.StatusCode))
```
zmb3: Can we simplify this error message?  How about: `spotify: HTTP %d: %s`?

spotify.go
```diff
@@ -107,14 +118,50 @@ func (e Error) Error() string {
...
+			}
+		}
+	} else if e.E.Error() == "" {
+		// Some errors will result in there being a useful status-code but an
+		// empty message, which will confuse the user (who only has access to
+		// the message and not the code). An example of this is when we send
+		// some of the arguments directly in the HTTP query and the URL ends-up
+		// being too long.
+
+		e.E.Message = fmt.Sprintf("spotify: there was an HTTP error (%d '%s') but we received an empty error", resp.StatusCode, http.StatusText(resp.StatusCode))
```
zmb3: Simplify this message too.  Something as simple as `spotify: unexpected HTTP %d: %s`

spotify.go
```diff
@@ -124,6 +171,78 @@ func decodeError(r io.Reader) error {
 // authenticate, you can use `DefaultClient`.
 type Client struct {
 	http *http.Client
+
+	autoRetry     bool
+	retryDuration time.Duration
+}
+
+func (c *Client) SetAutoRetry(flag bool) {
```
zmb3: Make `autoRetry` exported and remove the setter.  Users can set `client.AutoRetry = true` directly if they want to opt in to this behavior.

spotify.go
```diff
@@ -124,6 +171,78 @@ func decodeError(r io.Reader) error {
...
+	c.autoRetry = flag
+}
+
+// executeOpt executes a non-GET request. `needsStatus` describes another code
+// that can represent success. Note that in all current usages of this function,
+// we need to still allow a 200 even if we'd also like to check for a second
+// success code.
+func (c *Client) executeOpt(req *http.Request, needsStatus int, result interface{}) error {
```
zmb3: I think I would like to try a new pattern here.  Let's remove `executeOpt` and have a single `execute` method that looks like this:

```
func (c *Client) execute(req *http.Request, result interface{}, needsStatus ...int) error
```

Then inside execute, if `len(needsStatus) > 0` and the status code was not a 200 OK, you can check if the status code is found in the needsStatus slice.

The advantages to this are:
- only a single function that works with and without the optional status parameter(s)
- support for multiple valid status codes

spotify.go
```diff
@@ -124,6 +171,78 @@ func decodeError(r io.Reader) error {
...
+			return err
+		}
+
+		defer resp.Body.Close()
+
+		if resp.StatusCode == rateLimitExceededStatusCode && c.autoRetry {
+			time.Sleep(c.retryDuration)
+			continue
+		} else if resp.StatusCode != http.StatusOK {
+			errorMessage := decodeError(c, resp)
```
zmb3: Nitpick, but `decodeError` returns an `Error`, not a message, so let's not use `errorMessage` for a variable name.

In fact, just `return decodeError(c, resp)` would work fine.

zmb3:
Changes are looking great - I added a few comments on your revisions.

page.go
dsoprea: Done.

spotify.go
dsoprea: Done.

spotify.go
dsoprea: How about "spotify: HTTP %d: %s (body empty)"?

spotify.go
dsoprea: I'm not sure if "unexpected" is clear or accurate. How about "spotify: unexpected HTTP %d: %s (empty error)"?

spotify.go
dsoprea: Done.

spotify.go
dsoprea: Although technically this might be useful, there are only a handful of calls to `executeOpt` (mostly in playlist.go) and none of them check for more than one other success-code. Plus, we have many implementations of an "opt" function allowing for extended usage and a "non-opt" function allowing for a simpler, common usage. I believe that this function makes the most sense as is, but, if you insist, I'll make the change.

dsoprea:
Awaiting your responses. Let's get this merged.



Dustin

zmb3:
Hopping on a plane now. Will get back to you this evening!

spotify.go
zmb3: We do have many implementations of an "opt" function in the public API, but this is a private helper.  I'm not convinced the "opt" approach was the best decision anyway as it results in a pretty large API surface area, but that's what we're stuck with for now.

In any case, I don't have a strong preference either way.  I appreciate your contributions here, and if this makes the most sense to you lets keep it :+1:

dsoprea:
Thank you. Would you like to merge, or would you like to grant me write-access with which to merge it myself?

zmb3:
Everything looks pretty good to me.

Ideally I would love to see a test or two that exercises the new `decodeError` function (which is where the majority of the logic changes live), but I recognize you've spent a fair amount of time on this and I'm okay merging as is and going back to add the test later.

dsoprea:
Added two tests. Please merge (or give me the rights to).

zmb3:
Thanks again!

PR #40, opened 2017-05-21
Simplified NewReleasesOpt.

### LINKED ISSUE #38: NewReleasesOpt: Unnecessary complexity?
It seems like we're cutting `ContentLength`-bytes out of the body, loading it into a `Buffer`, and then passing that to the JSON-parser when we could just do what we did everywhere else and parse `resp.Body` directly. No?

Current:

```go
func (c *Client) NewReleasesOpt(opt *Options) (albums *SimpleAlbumPage, err error) {
    spotifyURL := baseAddress + "browse/new-releases"
    if opt != nil {
        v := url.Values{}
        if opt.Country != nil {
            v.Set("country", *opt.Country)
        }
        if opt.Limit != nil {
            v.Set("limit", strconv.Itoa(*opt.Limit))
        }
        if opt.Offset != nil {
            v.Set("offset", strconv.Itoa(*opt.Offset))
        }
        if params := v.Encode(); params != "" {
            spotifyURL += "?" + params
        }
    }
    resp, err := c.http.Get(spotifyURL)
    if err != nil {
        return nil, err
    }
    defer resp.Body.Close()
    if resp.StatusCode != http.StatusOK {
        return nil, decodeError(c, resp)
    }
    buf := bytes.NewBuffer(make([]byte, 0, resp.ContentLength+1))
    _, err = buf.ReadFrom(resp.Body)
    if err != nil {
        return nil, err
    }
    body := buf.Bytes()
    var objmap map[string]*json.RawMessage
    err = json.Unmarshal(body, &objmap)
    if err != nil {
        return nil, err
    }
    var result SimpleAlbumPage
    err = json.Unmarshal(*objmap["albums"], &result)
    if err != nil {
        return nil, err
    }
    return &result, nil
}
```

Proposed:

```go
func (c *Client) NewReleasesOpt(opt *Options) (albums *SimpleAlbumPage, err error) {
    spotifyURL := baseAddress + "browse/new-releases"
    if opt != nil {
        v := url.Values{}
        if opt.Country != nil {
            v.Set("country", *opt.Country)
        }
        if opt.Limit != nil {
            v.Set("limit", strconv.Itoa(*opt.Limit))
        }
        if opt.Offset != nil {
            v.Set("offset", strconv.Itoa(*opt.Offset))
        }
        if params := v.Encode(); params != "" {
            spotifyURL += "?" + params
        }
    }
    resp, err := c.http.Get(spotifyURL)
    if err != nil {
        return nil, err
    }
    defer resp.Body.Close()
    if resp.StatusCode != http.StatusOK {
        return nil, decodeError(c, resp)
    }
    var objmap map[string]*json.RawMessage
    err = json.NewDecoder(resp.Body).Decode(&objmap)
    if err != nil {
        return nil, err
    }

    var result SimpleAlbumPage
    err = json.Unmarshal(*objmap["albums"], &result)
    if err != nil {
        return nil, err
    }
    return &result, nil
}
```

dsoprea:
Closes #38.

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
-	"bytes"
 )
 
 // Version is the version of this library.
@@ -169,17 +168,12 @@ func (c *Client) NewReleasesOpt(opt *Options) (albums *SimpleAlbumPage, err erro
 	if resp.StatusCode != http.StatusOK {
 		return nil, decodeError(resp.Body)
 	}
-	buf := bytes.NewBuffer(make([]byte, 0, resp.ContentLength+1))
-	_, err = buf.ReadFrom(resp.Body)
-	if err != nil {
-		return nil, err
-	}
-	body := buf.Bytes()
 	var objmap map[string]*json.RawMessage
-	err = json.Unmarshal(body, &objmap)
+	err = json.NewDecoder(resp.Body).Decode(&objmap)
 	if err != nil {
 		return nil, err
 	}
+
 	var result SimpleAlbumPage
 	err = json.Unmarshal(*objmap["albums"], &result)
 	if err != nil {


zmb3:
Looks good, but I can't merge until the code is `gofmt`d.

`gofmt -w spotify.go` should do the trick, but I highly recommend setting up your editor to format on save.

dsoprea:
Done.

PR #43, opened 2017-05-28
Added get-recently-played.

dsoprea:


dsoprea:
Cancel. Suspect bad rebase.

PR #44, opened 2017-05-28
Added get-recently-played to player.

dsoprea:
Also added String() to `SimpleTrack`.

### DIFF
--- a/player.go
+++ b/player.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"time"
 )
 
 // PlayerDevice contains information about a device that a user can play music on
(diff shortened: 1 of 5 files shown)


player.go
```diff
@@ -62,6 +63,21 @@ type CurrentlyPlaying struct {
 	Item *FullTrack `json:"Item"`
 }
 
+type RecentlyPlayedItem struct {
+	// Track is the track information
+	Track SimpleTrack `json:"track"`
+
+	// PlayedAt is the time that this song was played
+	PlayedAt time.Time `json:"played_at"`
```
zmb3: Can we update the tests to make sure that `played_at` is unmarshaled correctly?  Other than that looks good to me.

dsoprea: Done.

PR #46, opened 2017-06-07
AddTracksToPlaylist POST URIs as body rather than query parameter

brodin:
Sending the URIs for the AddTracksToPlaylist function as body rather than query paramater improves performance when sending a large number of tracks (around 50-80 ish)

> A maximum of 100 tracks can be added in one request. Note: it is likely that passing a large number of track URIs as a query parameter will exceed the maximum length of the request URI. When adding a large number of tracks it is recommended to pass them in the request body

(https://developer.spotify.com/web-api/add-tracks-to-playlist/ )


zmb3:
Looks good, just need to rebase on latest master.

brodin:
@zmb3 done! :)

PR #50, opened 2017-07-14
ADD a function to get TOKEN without the request object

### LINKED ISSUE #49: Stop using http.Request for token recovery
Hello,

I'm actually using gin, and i can't access to the http.Request object
Is it possible to add a function where i can retrieve the token without using http.request

regards,

Preovaleo:
Fix #49 by adding a function to get the token without the request object

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -146,6 +146,13 @@ func (a Authenticator) Token(state string, r *http.Request) (*oauth2.Token, erro
 	return a.config.Exchange(a.context, code)
 }
 
+// TokenwithoutRequest pulls an authorization code from an HTTP request and attempts to exchange
+// it for an access token.  The standard use case is to call TokenwithoutRequest from the handler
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -146,6 +146,13 @@ func (a Authenticator) Token(state string, r *http.Request) (*oauth2.Token, erro
 	return a.config.Exchange(a.context, code)
 }
 
+// TokenwithoutRequest pulls an authorization code from an HTTP request and attempts to exchange
```
zmb3: Just a few comments:

1. Please update the godoc.  This doesn't _pull an authorization code from an HTTP request_ anymore.
2. I would prefer to still pass in the `state` and do a comparison internally.  It's too easy to forget otherwise and open yourself up to XSRF attacks.

strideynet:
Closing as abandoned.

PR #51, opened 2017-07-20
Refresh Token Feature

joaoaneto:
Feature to facilitate obtaining a new token from another already expired.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,1 @@
+build: off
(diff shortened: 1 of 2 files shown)


zmb3:
I don't think this should be necessary.  The Go OAuth package will refresh tokens transparently.

Additionally, you seem to have disabled appveyor builds.  Was this a mistake?

joaoaneto:
I will investigate the package better.
About appveyor builds, it was a mistake. Sorry!

PR #54, opened 2017-10-12
Use Go 1.8 and 1.9 for CI

adamo57:


zmb3:
Thanks for this.  Let's remove 1.7 and just keep 1.8 and 1.9.

adamo57:
Personally, I think it's better to test all of the previous go versions from when you started this project

zmb3:
That’s not necessary. This is a simple library. Go has a good compatibility promise.

adamo57:
Ok! Updated

PR #57, opened 2017-11-01
Changed Timestamp int to int64, could help the unmarshaling the JSON …

fhiden:
…object when calling the PlayerCurrentlyPlaying(). 

For me, i had a problem with having the Timestamp as int because of the size


fhiden:
ran it and it didnt change my problem :-1: 

im closing the pullrequest

zmb3:
Thanks for this.  Please add a test that fails without the change and passes with the change and this will be good to go :+1:

PR #58, opened 2017-11-13
SimpleAlbum with SimpleArtists

ptrkrlsrd:
# Added a slice of SimpleArtists to the SimpleAlbum struct

Hey, I've recently used this package for a project, and found that I needed the array of artists which the Spotify API returns when searching for albums. Therefore I decided to also create a pull request in case you're interested in implementing this change.

### DIFF
--- a/album.go
+++ b/album.go
@@ -13,7 +13,7 @@ import (
 type SimpleAlbum struct {
 	// The name of the album.
 	Name string `json:"name"`
-	// An array of SimpleArtists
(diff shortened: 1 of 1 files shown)


zmb3:
Thanks for this :-)

The term `array` can be misleading in Go, can we change the comment to read `slice` or `list`?

ptrkrlsrd:
I totally forgot that, but I’ll fix it as soon as possible :)

PR #61, opened 2018-03-06
Remove unnecessary pointers from structs

andrewpillar:
This is definitely going to be a breaking change for the people using this library. I ran the tests and they all passed after the changes I had made.

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -237,55 +237,55 @@ type Options struct {
 	// this parameter if you want the list of returned items to
 	// be relevant to a particular country.  If omitted, the
 	// results will be relevant to all countries.
-	Country *string
+	Country string
 	// Limit is the maximum number of items to return.
-	Limit *int
+	Limit int
 	// Offset is the index of the first item to return.  Use it
 	// with Limit to get the next set of items.
-	Offset *int
+	Offset int
 	// Timerange is the period of time from which to return results
 	// in certain API calls. The three options are the following string
 	// literals: "short", "medium", and "long"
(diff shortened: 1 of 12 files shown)


spotify.go
```diff
@@ -237,55 +237,55 @@ type Options struct {
...
+	if err := c.get(spotifyUrl, &objmap); err != nil {
 		return nil, err
 	}
 
-	var result SimpleAlbumPage
-	err = json.Unmarshal(*objmap["albums"], &result)
-	if err != nil {
+	result := new(SimpleAlbumPage)
+
+	if err := json.Unmarshal(*objmap["albums"], &result); err != nil {
```
zmb3: Does this actually work?  Why unmarshal into a **SimoleAlbumPage when a single * will do?

andrewpillar:
I think creating a new branch to merge this change into would be a smart thing to do. Then you could use gopkg.in.

spotify.go
andrewpillar: Gah, you're right, I forgot that `new` returns a pointer, so the `&` is unnecessary. Just an old habit where I always pass a pointer to `Unmarshal` by using `&` in front of the variable.

zmb3:
> I think creating a new branch to merge this change into would be a smart thing to do. Then you could use gopkg.in.

@andrewpillar can you update the pull request to merge into a new branch and not master?

andrewpillar:
@zmb3 sure I can, but I can currently only edit this pull request to merge into either `master` or `v1`.

PR #62, opened 2018-03-07
Added support for adding and removing albums from a user's library

ddevcap:


library.go
```diff
@@ -59,3 +59,57 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
...
+
+// UserHasAlbums checks if one or more albums are saved to the current user's
+// "Your Music" library.
+func (c *Client) UserHasAlbums(ids ...ID) ([]bool, error) {
+	if l := len(ids); l == 0 || l > 50 {
+		return nil, errors.New("spotify: UserHasAlbums supports 1 to 50 IDs per call")
+	}
+	spotifyURL := fmt.Sprintf("%sme/albums/contains?ids=%s", c.baseURL, strings.Join(toStringSlice(ids), ","))
+
+	var result []bool
```
zmb3: In many cases, the final `err` check can be omitted.  For example:
```
var result []bool
err := c.get(spotifyURL, &result)
return result, err
```

library.go
```diff
@@ -59,3 +59,57 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
...
+	spotifyURL := fmt.Sprintf("%sme/albums?ids=%s", c.baseURL, strings.Join(toStringSlice(ids), ","))
+	method := "DELETE"
+	if add {
+		method = "PUT"
+	}
+	req, err := http.NewRequest(method, spotifyURL, nil)
+	if err != nil {
+		return err
+	}
+	err = c.execute(req, nil)
```
zmb3: Similarly: `return c.execute(req, nil)`

library_test.go
```diff
@@ -55,3 +55,54 @@ func TestRemoveTracksFromLibrary(t *testing.T) {
...
+		t.Error("Expected [false, true], got", contains)
+	}
+}
+
+func TestAddAlbumsToLibrary(t *testing.T) {
+	client, server := testClientString(http.StatusOK, "")
+	defer server.Close()
+
+	err := client.AddAlbumsToLibrary("47DiM7xqFvwplamrB1GQJo", "73TmwDD6mBOZh6sF9sKXZo")
+	if err != nil {
```
zmb3: This doesn't test much other than that we return a nil error.

I wonder if we can make some assertions to verify that the correct endpoint was hit with the correct data.

library.go
ddevcap: You're right! I've made some changes thruout the codebase.

ddevcap:
Updated the album_test. This was the only test that i could find where it would fail if the requested object wasn't a nil pointer. All other test are looking at the err. So i removed that assertion to make it behave like the other tests. Err should always be checked before continuing, right? What do you think?

library_test.go
ddevcap: True, i'll have a crack at this one when i'm back from ski holiday ;) ⛷🎉

ddevcap:
Or revert those changes and update the tests ;)

a-schuurman:
Nice PR! I need this too :) Do you guys have an update on this?

zmb3:
Sorry for losing track of this.

I wasn't super clear and for that I'm sorry.
In my original comment, I suggested changes like the following:

```
var result []bool
err := c.get(spotifyURL, &result)
return result, err
```
or
```
return c.execute(req, nil)
```
The key here, especially in the first example is that in the case of an error, `result` should still be `nil`.  I think that behavior is important to preserve, as most APIs behave this way, and ones that don't typically document it explicitly.

I'm fine with either backing out to your first commit and merging that, or having you go back through and re-evaluate the error checking that was unrelated to the original feature. What do you think?

ddevcap:
I see! It would indeed be a breaking change if we return an empty struct. Altho it is a general good practice to check for errors before using the returned object. Also, if you where not to check for the error you would then get a runtime error because you're trying to access a property that does not exists on the object. So returning an empty struct would also prevent that right? 

But i completely see your point. I'll revert the changes and commit as you mentioned: "go back through and re-evaluate the error checking"

strideynet:
Closing as exceptionally behind master.

PR #67, opened 2018-07-01
Change type of userID to string, to be concistent with other code

eanker:
The ownerID is of type string, except for those three lines.

This caused errors when doing `user.ID` as it resulted in expected type ID got type string.

### DIFF
--- a/playlist.go
+++ b/playlist.go
@@ -128,7 +128,7 @@ func (c *Client) FeaturedPlaylists() (message string, playlists *SimplePlaylistP
 // user's public playlists.  To be able to follow playlists privately, the user
 // must have granted the ScopePlaylistModifyPrivate scope.  The
 // ScopePlaylistModifyPublic scope is required to follow playlists publicly.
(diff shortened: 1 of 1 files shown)


zmb3:
This is intentional, as it helps to distinguish cases where the user is just your username (a string) and where it is the “Spotify ID” which is something different.

Further, a change like this would break the public API so unless there’s very good reason to do so we should try to avoid it.

eanker:
Ah, thanks for the reply, didn't think of that. 

How would one then use on of those methods, to for example unfollow it's own playlist. 
As when I do `client.getCurrentUser().ID` I get a string, that can not be put in the method as ownerID.
And when having a playlist `client.UnfollowPlaylist(playlist.Owner.ID, playlist.ID)` also gives an error that the owner is an string opposed to a ID.

zmb3:
You can always do 

spotify.ID(playlist.Owner.ID)

eanker:
This change was not in line with the spotify API itself, and therefor is closed.

Issue #77, opened 2018-09-13
Changes to Playlist URIs coming to our APIs


brodin:
The day is here https://developer.spotify.com/community/news/2018/06/12/changes-to-playlist-uris/

This client is really hard to use now if you need to work with playlists without the user part. (eg `spotify:playlist:37i9dQZF1DZ06evO3OC4Te`). 

Can we make a communal effort to add handlers for Playlists operations only using the ID part?

PR #78, opened 2018-09-13
Change the Playlist API to support user less URIs

brodin:
Breaking change to support https://developer.spotify.com/community/news/2018/06/12/changes-to-playlist-uris/ 

The idea of introducing alternative paths to be not break the API occurred to me, but it felt like bloating the interface completely. Anyone upgrading to this version should without effort be able to migrate to the new interface.

Fixes https://github.com/zmb3/spotify/issues/77


zmb3:
Thanks for this, looks good!

PR #81, opened 2018-09-27
No content response don't return io.EOF error

### LINKED ISSUE #72: PlayerCurrentlyPlaying returns err io.EOF when not playing
Sometimes, when there's no player registered (I suppose), the API returns an EOF. The token _is_ correct and the error can be ignored.

```golang
	if ftl, err := client.PlayerCurrentlyPlaying(); err != nil {
		if err != io.EOF {
			log.Fatal(err)
		}
	} else {
		if !ftl.Playing {
			log.Println("Currently not playing...")
			return
		}
		// Do whatever
	}
```

chrisvdg:
* execute doesn't parse on 204 no content http status
* PlayerCurrentlyPlaying returns ErrNoPlayer when 204 is returned

closes #72

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -179,7 +179,7 @@ func (c *Client) execute(req *http.Request, result interface{}, needsStatus ...i
 			return c.decodeError(resp)
 		}
 
-		if result != nil {
+		if result != nil && resp.StatusCode != http.StatusNoContent {
 			if err := json.NewDecoder(resp.Body).Decode(result); err != nil {
 				return err
 			}


player.go
```diff
@@ -3,12 +3,16 @@ package spotify
...
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/url"
 	"strconv"
 	"time"
 )
 
+// ErrNoPlayer is an error returned when spotify could not find a connected player
+var ErrNoPlayer = errors.New("spotify: no player connected")
```
zmb3: I'm glad we're not returning an EOF anymore, but I'm not sure I like the idea of returning an error for a 204 status code.

204 is a successful code.  Spotify does not see this as an error, so why should this API?

chrisvdg: Hmm true, I was thinking of making it easy for the user to check no player was returned.
You'r rather want to return Nil or an empty struct and let users figure out themselves it was and empty response?

Issue #94, opened 2019-06-09
show_dialog optional parameter for authentication


linux-jedi:
The Spotify Web API also supports the option parameter `show_dialog` for authentication. Right now the library does not support this parameter and it's something that I would like to use for a project. Is there any plans to add/support this?

zmb3:
I’m happy to take a PR to add this.

PR #95, opened 2019-06-11
Add AuthURLWithDialog to force app authentication with Spotify

linux-jedi:
#94 Added a new showDialog boolean to the `Authenticator` struct and a setter method for it. For now, it is initialized to false by default. I didn't want to break the existing API/flow, so I didn't add it to the constructor or existing configuration methods. I would love any feedback/suggestions/changes as I am not super familiar with Go code bases :)

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -131,6 +131,11 @@ func (a Authenticator) AuthURL(state string) string {
 	return a.config.AuthCodeURL(state)
 }
 
+// AuthURLWithDialog returns the same URL as AuthURL, but sets show_dialog to true
+func (a Authenticator) AuthURLWithDialog(state string) string {
+	return a.config.AuthCodeURL(state, oauth2.SetAuthURLParam("show_dialog", "true"))
+}
+
 // Token pulls an authorization code from an HTTP request and attempts to exchange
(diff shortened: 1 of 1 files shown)


zmb3:
I would probably not worry about breaking the API and would instead make `showDialog` an argument to `AuthURL`.

linux-jedi:
Can we merge these changes?

zmb3:
Sorry @linux-jedi - didn't see your update.  Just one small request - can we _not_ export the show dialog constant?  (I'm not convinced we need a constant at all since there's only one use, but if we must have it, lets at least keep it internal to the package.)

linux-jedi:
I went ahead and just removed the constant. I agree – it is only being used once, so there's no real need for it.

zmb3:
Nice, thanks. I think we can remove the show dialog Boolean arg as well since there’s no need to have the toggle there. If you don’t want the dialog you just call the original method instead.

linux-jedi:
Nice catch. Totally was not thinking when I changed it from my original approach.

zmb3:
Thanks @linux-jedi!  Sorry for the delay.

PR #98, opened 2019-07-30
Add more fields to SimpleAlbum, SimpleTrack from api

rociiu:
Screen Shot 2019-07-30 at 11 38 50 PM

### DIFF
--- a/track.go
+++ b/track.go
@@ -33,9 +33,10 @@ type SimpleTrack struct {
 	// The number of the track.  If an album has several
 	// discs, the track number is the number on the specified
 	// DiscNumber.
-	TrackNumber int `json:"track_number"`
-	URI         URI `json:"uri"`
-	Popularity  int `json:"popularity"`
+	TrackNumber int               `json:"track_number"`
+	URI         URI               `json:"uri"`
+	Popularity  int               `json:"popularity"`
+	ExternalIDs map[string]string `json:"external_ids"`
 }
 
 func (st SimpleTrack) String() string {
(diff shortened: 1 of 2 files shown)


track.go
```diff
@@ -33,6 +35,8 @@ type SimpleTrack struct {
 	// DiscNumber.
 	TrackNumber int `json:"track_number"`
 	URI         URI `json:"uri"`
+	Popularity  int `json:"popularity"`
+        ExternalIDs map[string]string `json:"external_ids"`
```
zmb3: Looks like you need to run `gofmt`

zmb3:
I think this change is incorrect.  I don't see any of these changes documented in the object model:
- https://developer.spotify.com/documentation/web-api/reference/object-model/#track-object-simplified
- https://developer.spotify.com/documentation/web-api/reference/object-model/#album-object-simplified

Here's the [interesting part](https://developer.spotify.com/documentation/web-api/reference/player/get-recently-played/): 
> On success, the HTTP status code in the response header is 200 OK and the response body contains an array of play history objects (wrapped in a cursor-based paging object) in JSON format. The play history items each contain the context the track was played from (e.g. playlist, album), the date and time the track was played, and a track object (simplified). On error, the header status code is an error code and the response body contains an error object. If private session is enabled the response will be a 204 NO CONTENT with an empty payload.

So I think what we need to do is add a new `PlayHistory` type and not touch `SimpleTrack` or `SimpleAlbum`

rociiu:
@zmb3 i've corrected the format, also submitted a issue in https://github.com/spotify/web-api/issues/1314 .

zmb3:
Thanks @rociiu - let's see what they say on the upstream issue.

strideynet:
Closing as abandoned.

PR #100, opened 2019-08-02
extend expected http.StatusAccepted with http.StatusNoContent (closes #99)

### LINKED ISSUE #99: Using spotifyd
Hello there!
First of great work and thanks for making this package! :+1: :heart: 

I am trying this package to control [spotifyd](https://github.com/Spotifyd/spotifyd), however, somehow spotify API always returns with `HTTP 202: Accepted (body empty)` instead of `HTTP 204: No Content` as soon as I make a call (selected spotifyd as connected device).
The commands get trough! And spotifyd does everything fine, but `HTTP 202` is not expected in this package so it will aways return an error.

I know that there is an Autoretry mode, however spotify always returns 202 no matter what, so I end up in an endless loop...

My guess is that there is an issue on spotifyd's side. (Will create a ticket there as well)
However I think there should be an option to enable `HTTP 202 is good enough` mode

What do you think?
BTW: willing to implement that feature!

Eun:
closes #99

### DIFF
--- a/player.go
+++ b/player.go
@@ -202,7 +202,7 @@ func (c *Client) PlayerCurrentlyPlayingOpt(opt *Options) (*CurrentlyPlaying, err
 	}
 
 	var result CurrentlyPlaying
-	err = c.execute(req, &result, http.StatusNoContent)
+	err = c.execute(req, &result, http.StatusNoContent, http.StatusAccepted)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


Eun:
Will fix these tests asap.

Eun:
Ready to merge

zmb3:
Can we just make `execute` smart enough to treat all 2XX codes as successful so we don't have to pass `http.StatusAccepted` to every call?

Eun:
Accepting all 200 codes feels a bit risky, instead I would rather just add the specified 200 codes from spotify:

https://developer.spotify.com/documentation/web-api/#response-status-codes

zmb3:
Sure, but the reason we're in this situation is that Spotify sometimes sends 200 codes that it doesn't document, and thus this library issues an error when the operation succeeded..

Eun:
I haven't seen 200 codes that are undocumented. I can extend `execute` to treat 200, 201, 202 and 204 as success like they say [here](https://developer.spotify.com/documentation/web-api/#response-status-codes)

zmb3:
@Eun any follow up on extending `execute` as mentioned above?

PR #102, opened 2019-10-01
Move ReleaseDateTime method to SimpleAlbum struct

### LINKED ISSUE #101: ReleaseDateTime method only defined for FullAlbum struct
#71, which adds the `ReleaseDate` and `ReleaseDatePrecision` fields to the `SimpleAlbum` struct, was merged ages ago, but `SimpleAlbum` doesn't have the `ReleaseDateTime` method which is defined for the `FullAlbum` struct.

Is it possible to move the `ReleaseDateTime` method to the `SimpleAlbum` instead?
I would gladly open a PR if interested.

Tangentially related: Is there a reason why - especially in case of the Album structs - multiple fields are defined both on the "simple" as well as on the "full" variant even though the simple one is embedded?

siku2:
Resolves #101 by moving the `ReleaseDateTime` method to the `SimpleAlbum` struct and removing the unnecessary re-declarations of various fields from the `FullAlbum` struct.

### DIFF
--- a/album.go
+++ b/album.go
@@ -62,22 +62,14 @@ type Copyright struct {
 // FullAlbum provides extra album data in addition to the data provided by SimpleAlbum.
 type FullAlbum struct {
 	SimpleAlbum
-	Artists    []SimpleArtist `json:"artists"`
-	Copyrights []Copyright    `json:"copyrights"`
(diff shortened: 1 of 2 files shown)


zmb3:
Thanks, Simon!

PR #103, opened 2019-10-08
Add unit test example for clientcreds

jonatasbaldin:
Hi there 👋 

This PR adds an example on how to mock the `spotify.Client` for testing with a full description on a blog post I wrote.

Hope it helps people starting with the library ✨


zmb3:
Thanks for taking the time to send this PR, and to write up that blog post.  Good stuff.

The mocking example you provide is a great strategy for testing any API client, but I don’t see anything specific to this library that would require it to live in the codebase, so I’m going to close this.

Thanks again :+1:

jonatasbaldin:
Alright, thanks for the reply 😄 the library is great!

PR #107, opened 2019-11-19
Add New(...) constructor for client

jchorl:
Hi!

I'd like to instantiate my own Spotify client. Following [this](https://developers.google.com/calendar/quickstart/go) Google Oauth example, I want to do something like
```golang
	httpClient, err := util.GetOauthClient(context.TODO(), cfg, "spotifycreds.json")
	if err != nil {
		return spotify.Client{}, fmt.Errorf("GetOauthClient(): %w", err)
	}

	return spotify.New(spotify.WithHTTPClient(httpClient)), nil
```

I've written generic boilerplate for generating an oauth client that I can use for multiple libs, like google's calendar lib ([here](https://gist.github.com/jchorl/f5933485593e6cd1ab1450a0e607482f))

I opted to follow the functional options pattern described [here](https://github.com/tmrts/go-patterns/blob/f978e420361704bd7531e2b57905a308a3a012c8/idiom/functional-options.md) and also modified the tests to test out the new constructor.

The docs are updated to still recommend the standard auth flow.

Let me know what you think!

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -162,10 +162,7 @@ func (a Authenticator) Exchange(code string) (*oauth2.Token, error) {
 // NewClient creates a Client that will use the specified access token for its API requests.
 func (a Authenticator) NewClient(token *oauth2.Token) Client {
 	client := a.config.Client(a.context, token)
-	return Client{
-		http:    client,
-		baseURL: baseAddress,
-	}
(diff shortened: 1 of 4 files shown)


zmb3:
Closing since #123 was merged.

PR #110, opened 2019-12-13
Client.execute: Don't return EOF for response status 204

### LINKED ISSUE #109: PlayerCurrentlyPlaying() returns error EOF when nothing is playing
Hi there! 

I noticed that the function `client.PlayerCurrentlyPlaying()` returns an error with the message EOF when no song is currently playing.

The error occurs when [trying to decode an empty response body](https://github.com/zmb3/spotify/blob/869e03d/spotify.go#L184).
```go
if result != nil {
    if err := json.NewDecoder(resp.Body).Decode(result); err != nil {
        return err
    }
}
```
Debugger Screenshot

I would offer to submit a PR to fix this, but this is actually the first project I'm working on using go, so I'm not sure how to approach this issue.

Thanks!

Nikos410:
Implements the behaviour from #88 for `client.execute()` as well.

This prevents an EOF error when client.execute() receives status 204 (No Content), e.g. when `client.PlayerCurrentlyPlaying()` is called while nothing is playing.

Resolves #109.

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -174,6 +174,9 @@ func (c *Client) execute(req *http.Request, result interface{}, needsStatus ...i
 			time.Sleep(retryDuration(resp))
 			continue
 		}
+		if resp.StatusCode == http.StatusNoContent {
+			return nil
+		}
 		if (resp.StatusCode >= 300 ||
 			resp.StatusCode < 200) &&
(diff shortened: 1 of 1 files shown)


zmb3:
Thanks @Nikos410!!

PR #115, opened 2020-03-30
feat: Enable option to not sent market using GetArtistAlbumsOpt

brodin:
https://github.com/zmb3/spotify/blob/master/artist.go#L127-L131

I have a use case where I want ALL albums for an artist and then I need to be able to not send the `markets` url-value. 
Edit: ~This is my suggestion on how to do it while keeping the current behaviour where an unspecified market will give you US as default.~

Suggestions for other implementations are welcome 🙏

### DIFF
--- a/artist.go
+++ b/artist.go
@@ -112,7 +112,9 @@ func (c *Client) GetArtistAlbums(artistID ID) (*SimpleAlbumPage, error) {
 // parameters used to filter and sort the result.
 //
 // The AlbumType argument can be used to find a particular type of album.  Search
-// for multiple types by OR-ing the types together.
+// for multiple types by OR-ing the types together. If the market (Options) is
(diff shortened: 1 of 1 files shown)


zmb3:
Seems reasonable to me.  Can we just document this behavior on the method docs?

brodin:
> Seems reasonable to me. Can we just document this behavior on the method docs?

Of course, the is fallback is however not documented. The more I read the less confused I get. If you pass in `nil` as options you *don't* get the `options.Country = US` default. I am thinking that we should simplify remove it and add the disclaimer(below) in the method documentation instead?   

```
// if the market is not specified, Spotify will likely return a lot
// of duplicates (one for each market in which the album is available)
// - prevent this behavior by falling back to the US by default
```

brodin:
@zmb3  I force-pushed a new approach – what do you think?

PR #116, opened 2020-03-31
GetArtistAlbumsOpt: Improvements and bug fixes

brodin:
This API will be backwards compatible, but also support the full potential of the underlying endpoint. See https://developer.spotify.com/documentation/web-api/reference/artists/get-artists-albums/ for reference

### DIFF
--- a/artist.go
+++ b/artist.go
@@ -123,7 +123,7 @@ func (c *Client) GetArtistAlbumsOpt(artistID ID, options *Options, ts ...*AlbumT
 		for i := range ts {
 			types[i] = ts[i].encode()
 		}
-		values.Set("album_type", strings.Join(types, ","))
+		values.Set("include_groups", strings.Join(types, ","))
 	}
 	if options != nil {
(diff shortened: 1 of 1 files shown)


zmb3:
Thanks!

brodin:
@zmb3 I found an issue with this.
If you explicitly send in `nil` as album type this will crash. This since `ts` then will have length 1 and this loop will result in a nil-pointer https://github.com/zmb3/spotify/pull/116/files#diff-234075626fc63bfff8ccab16d40509e9R124 

So the code is not fully backwards compatible. Do you want me to add a guard for it?

The guard would be to change this line https://github.com/zmb3/spotify/pull/116/files#diff-234075626fc63bfff8ccab16d40509e9R121 to

`if ts != nil && ts[0] != nil` then we should be fully backwards compatible.

zmb3:
@brodin sure.  Note that you don't even need the `if ts != nil` check, as you can range over a nil slice, and the loop simply won't execute.

PR #122, opened 2020-04-19
Remove dependency on env

pixelrazor:
NewAuthenticator shouldn't depend on the os environment; how those are supplied should be up to the user.

Removed the method to explicitly set those values. NewAuthenticator shouldn't return an invalid object that has to be configured with another function

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -4,10 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"net/http"
-	"os"
-
 	"golang.org/x/oauth2"
+	"net/http"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


zmb3:
I get your point, but this is largely subjective and breaks existing users.  If we’re going to break things for people and cause them extra work to upgrade, there should be a really good reason.  Is there something that’s not possible in the current implementation that this PR enables?

pixelrazor:
> 
> 
> I get your point, but this is largely subjective and breaks existing users. If we’re going to break things for people and cause them extra work to upgrade, there should be a really good reason. Is there something that’s not possible in the current implementation that this PR enables?

Nope, i wasn''t expecting a merge. I just wanted to put thoughts out for consideration if there's ever a v2

PR #123, opened 2020-04-22
Allow client creation from authorizing HTTP client

CodyDWJones:
This change allows apps to supply a http.Client they already have, rather than relying on the Authenticator helper. An app might use common code to obtain an auth'd http.Client because it interacts with many OAuth services.

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -41,12 +41,22 @@ const (
 const baseAddress = "https://api.spotify.com/v1/"
 
 // Client is a client for working with the Spotify Web API.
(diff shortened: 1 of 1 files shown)


CodyDWJones:
I didn't realize #107 had been submitted. Either approach satisfies my requirement. #107 is more future proof (if a non-exported field is added to Client) but until then quite a bit more verbose.

zmb3:
Thanks for this, #107 is probably a nicer approach, but this is 100% backwards compatible so I'll merge this and we'll keep #107 around in case somebody wants to work on a v2 of this library.

PR #125, opened 2020-05-10
Implement changes in the audio analysis endpoint response

### LINKED ISSUE #124: Timestamp parsing of analysis meta fails
The type of the Timestamp value in the audio-analysis endpoint reponse changed from int to string. Thereby the json parsing in the following line is failing with the error message `json: cannot unmarshal string into Go struct field AnalysisMeta.meta.timestamp of type int64`. https://github.com/zmb3/spotify/blob/5fe5f9535a2c486f0e57ddd88bb703e297329edb/audio_analysis.go#L34

Response from `https://api.spotify.com/v1/audio-analysis/{id}`:
```
{
  "meta": {
    "analyzer_version": "4.0.0",
    "platform": "Linux",
    "detailed_status": "OK",
    "status_code": 0,
    "timestamp": "1456010389",
    "analysis_time": 9.1394,
    "input_process": "libvorbisfile L+R 44100->22050"
  },
```

Because this change is nowhere documented im unsure if this will be reverted. Should the library support both cases?

qnten:
Closes #124 and additionally implements changes of the `num_samples` and `analysis_sample_rate` fields.

### DIFF
--- a/audio_analysis.go
+++ b/audio_analysis.go
@@ -71,7 +71,7 @@ type AnalysisTrack struct {
 	SampleMD5               string  `json:"sample_md5"`
 	OffsetSeconds           int     `json:"offset_seconds"`
 	WindowSeconds           int     `json:"window_seconds"`
-	AnalysisSampleRate      int64   `json:"analysis_sample_rate"`
+	AnalysisSampleRate      float64 `json:"analysis_sample_rate"`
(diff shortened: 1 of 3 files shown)


zmb3:
Are these changes backwards compatible, or does it now require a string (meaning it will break with number)?  

As a side note - I never knew about `,string` in a JSON tag.  Thanks for teaching me something :-)

PR #126, opened 2020-05-21
added: CurrentUsersShows(), CurrentUsersShowsOpt()

pebbe:
Added support for user's saved shows (podcasts)


rexroof:
Any possibility to get standalone show metadata and/or return a list of a show's episodes? 
I'd work on my own PR but I'm only just learning go.

PR #127, opened 2020-05-24
feat(user): save/remove albums and tracks

Easen:
Added 4 new methods
- SaveAlbum
- RemoveAlbum
- SaveTrack
- RemoveTrack

Refactor - Modified `modifyResource` to accept a resource type

### DIFF
--- a/user.go
+++ b/user.go
@@ -229,32 +229,20 @@ func (c *Client) CurrentUserFollows(t string, ids ...ID) ([]bool, error) {
 	return result, nil
 }
 
-type userResourceType uint
-
 const (
-	followingResource userResourceType = iota
-	albumsResource
-	tracksResource
+	followingResource = "following"
+	albumsResource    = "albums"
+	tracksResource    = "tracks"
 )
 
-var userResourceTypeStrings = []string{
-	"following",
-	"albums",
-	"tracks",
-}
-
-func (urt userResourceType) encode() string {
-	return userResourceTypeStrings[urt]
-}
-
-func (c *Client) modifyResource(resource userResourceType, usertype string, follow bool, ids ...ID) error {
+func (c *Client) modifyResource(resourceType string, usertype string, follow bool, ids ...ID) error {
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -193,14 +229,32 @@ func (c *Client) CurrentUserFollows(t string, ids ...ID) ([]bool, error) {
 	return result, nil
 }
 
-func (c *Client) modifyFollowers(usertype string, follow bool, ids ...ID) error {
+type userResourceType uint
+
+const (
+	followingResource userResourceType = iota
```
zmb3, line 235: This PR looks great, thanks!

The only thing I'd change is these constants. As written, it's easy to make a mistake and forget to keep the resource constant in sync with `userResourceTypeStrings`.

Can we just make the constant a `string` instead of a `uint`?  Then there would be no need for a constant and a slice to lookup the string.

Easen: Sure thing, I supposed it doesn't need to be strongly typed as it's a private method.

user.go
```diff
@@ -193,14 +229,20 @@ func (c *Client) CurrentUserFollows(t string, ids ...ID) ([]bool, error) {
...
+	albumsResource    = "albums"
+	tracksResource    = "tracks"
+)
+
+func (c *Client) modifyResource(resourceType string, usertype string, follow bool, ids ...ID) error {
 	if l := len(ids); l == 0 || l > 50 {
 		return errors.New("spotify: Follow/Unfollow supports 1 to 50 IDs")
 	}
 	v := url.Values{}
 	v.Add("type", usertype)
```
zmb3, line 243: It looks like the `type` parameter is only necessary for updating followers.

In addition, the error we issue if too many IDs are provided is specific to follow/unfollow.

Its starting to seem like we should have two separate methods - one for modifying followers, and one generic one for library manipulation (saving tracks, albums, shows, etc.)

strideynet:
Closing as abandoned

PR #128, opened 2020-05-31
add support for a user's saved songs

erooke:
Added the functions `UserSavedTracks` and `UserSavedTracksOpt` which implement the api endpoint documented [here](https://developer.spotify.com/documentation/web-api/reference/library/get-users-saved-tracks/). Also threw in an example of it being used.

### DIFF
--- a/library.go
+++ b/library.go
@@ -59,3 +59,32 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
 	}
 	return nil
 }
+
+// UserSavedTracks returns the tracks which the current user has saved. This 
+// call requires the ScopeUserLibraryRead scope. Limit specifies the max items
+// per page.
+func (c *Client) UserSavedTracks(limit int) (SavedTrackPage, error) {
+  return c.UserSavedTracksOpt(limit, 0, "")
+}
+
+// UserSavedTracksOpt returns tracks which the current user has saved. This
+// call requires the ScopeUserLibraryRead scope. Limit and offset are used for
+// manual paging. Limit is the max number of objects to return in the page,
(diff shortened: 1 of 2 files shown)


library.go
```diff
@@ -59,3 +59,32 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
...
+  return c.UserSavedTracksOpt(limit, 0, "")
+}
+
+// UserSavedTracksOpt returns tracks which the current user has saved. This
+// call requires the ScopeUserLibraryRead scope. Limit and offset are used for
+// manual paging. Limit is the max number of objects to return in the page,
+// offset is the index of the first object in the page. Market is an ISO 3166-1
+// alpha-2 country code, or the string "from_token". This is used for track
+// relinking, use the empty string to leave this unspecified.
+func (c *Client) UserSavedTracksOpt(limit int, offset uint, market string)(SavedTrackPage, error) {
```
zmb3, line 76: It looks like this code may need to be formatted with `gofmt`.

zmb3, line 76: Can we leverage the `spotify.Options` type to specify the limit and offset?

library.go
```diff
@@ -59,3 +59,32 @@ func (c *Client) modifyLibraryTracks(add bool, ids ...ID) error {
...
+  var result SavedTrackPage
+  if limit > 50 || limit < 1 {
+    return result, errors.New("spotify: SavedTracks supports a limit of 1 to 50")
+  }
+  spotifyURL := fmt.Sprintf("%sme/tracks?limit=%d", c.baseURL, limit)
```
zmb3, line 81: Use `url.Values` instead of building your own URL with `Sprintf`.

strideynet:
Closing as abandoned

Issue #108, opened 2019-12-09
How to refresh access token?


Nikos410:
Hi there and thanks for this great library, it saved us a lot of work :)

I do have one problem though - an access token expires after 1 hour, but can be refreshed using the refresh token:

> Access tokens are deliberately set to expire after a short time, after which new tokens may be granted by supplying the refresh token originally obtained during the authorization code exchange.

(https://developer.spotify.com/documentation/general/guides/authorization-guide#4-requesting-a-refreshed-access-token-spotify-returns-a-new-access-token-to-your-app)

Is there a way to do this using this library?

Thanks in advance!

zmb3:
See also #7.  If you're using the authorization code flow, the library should refresh the token for you automatically.

If you're using the client credentials flow folks have just been recreating the client.  Definitely open to ideas for making that workflow easier.

chew-z:
As per [this discussion](https://github.com/golang/oauth2/issues/84#issuecomment-520099526) `newToken, err := src.Token() // this actually goes and renews the tokens` which implies, I think, that zmb3/spotify/auth.go `client.Token()` should also refresh token.

Now, this is just sidenote but you rarely need spotify token to last beyond default 60 minutes.

1. If you store spotify cookies you would be re-authorized transparently in the background
2. oauth2 token had been designed to be stateful not stateless (hence variable called state). Many people for simplicity code per app token (like `const state=abc123`) but it is note as intended.

Nikos410:
Hi, sorry for taking a while to respond.

I think I'm using the authorization code flow, like in the example in README.md.

So that means, if I create a client with `Authenticator.NewClient(token)` and then perform actions using that client, the token will be updated automatically?

chew-z:
Yes. Your client should be authorized even if more then an hour passed.

Do `log.Println(token.Expiry.Sub(time.Now()))` to make sure.
And `newToken, _ := client.Token()` to obtain new token implicit.

If you would like to store new token (to file or database) this pseudocode might help.

```
		tok, err := getTokenFromDB()
		if err == nil {
			client := auth.NewClient(tok)

				if m, _ := time.ParseDuration("5m30s"); time.Until(tok.Expiry) < m {
					newToken, _ := client.Token()
					updateTokenInDB(newToken)
				}
			return &client
		}
```

Nikos410:
Alright, thank you very much, that does help me a lot

Issue #7, opened 2016-02-29
Spotify refresh token after 1 hour expiry


jesseward:
Realize this is not so much an issue with this Spotify lib, but curious if anyone has got  Spotify's OAUTH token to "refresh" using the Go Ouath2 library?

I have implemented a pair of helper functions to writeToken and readToken that caches a json blob containing the auth token to disk. I am doing this so that the user isn't forced to continually authorize access to Spotify. As it stands Spotify has a very short token expiry (1 hour) but they do provide a refresh token when authenticated.  Documented here -> https://developer.spotify.com/web-api/authorization-guide/#authorization_code_flow

@zmb3 if you have attempted such an oauth flow using the refresh token, would you mind adding this to the example?

zmb3:
I haven't played with this in detail, but it is my understanding that the token should refresh automatically.  See https://godoc.org/golang.org/x/oauth2#Config.Client:

>Client returns an HTTP client using the provided token. The token will auto-refresh as necessary.

If you're seeing something different or have suggestions for improving the authentication portion I'm all ears :smile:

zmb3:
I put together a really simple app that authenticates and makes a call that requires authentication every 10 minutes.  I let it run for several hours and there have been no issues, so I believe the token is refreshing as expected.

avinassh:
Can you tell me what I am doing wrong? I am saving all the details to a file first and then reading from that. Spotify tokens expire after an hour and I guess refresh isn't happening. Here is my code:

```
	token := &oauth2.Token{
		Expiry:       time.Unix(config.TokenExpiry, 0),
		TokenType:    config.TokenType,
		AccessToken:  config.AccessToken,
		RefreshToken: config.RefreshToken,
	}
	fmt.Println(token.Valid()) // this prints false after an hour and code starts failing
	client := auth.NewClient(token)
	user, err := client.CurrentUser()
	if err != nil {
		log.Fatal("Failed :", err)
	}
	fmt.Println("Hi: ", user.ID)
```

And I get an error:

```
2017/03/19 22:11:43 Failed :Get https://api.spotify.com/v1/me: oauth2: cannot fetch token: 400 Bad Request
```

To produce this error, fetch tokens from Spotify, save them and run the script again after an hour.

jordanpotter:
[...]

PR #137, opened 2020-10-13
Extend authcode example to include setting client-id, client secret and reusing saved token

fabstu:
There are multiple issues showing that it is not clear that client-id and client-secret are needed for the refresh. This PR makes it clear in the hopes we see less issues.

- https://github.com/zmb3/spotify/issues/7
- https://github.com/zmb3/spotify/issues/108

### DIFF
--- a/examples/authenticate/authcode/authenticate.go
+++ b/examples/authenticate/authcode/authenticate.go
@@ -11,8 +11,10 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os"
 
 	"github.com/zmb3/spotify"
+	"golang.org/x/oauth2"
 )
 
 // redirectURI is the OAuth redirect URI for the application.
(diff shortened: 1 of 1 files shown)


zmb3:
You shouldn't need to set the client ID and secret if you're using the default environment variables.

https://github.com/zmb3/spotify/blob/9bec46940cc04f095544e94c0cbc79e7772fd61d/auth.go#L97-L98

fabstu:
The point here is for people using different variables or loading from different places. They have to use ˋSetAuthInfoˋ for the first time and for the refresh. This was not clear to me. This PR is to make this crystal clear. I know this now so I do not care much but new users might stumble here which, I feel, is unnecessary.

strideynet:
Closing as abandoned for now, plus the API has changed significantly.

PR #138, opened 2020-11-15
Fixing CurrentUsersTracksOpt market query parameter + proper handling of results

fclairamb:
Thank you for this great lib. 

- The `me/tracks` API uses the `market` query parameter instead of `country`.
- When the `market` query parameter is passed, an `is_playable` and `linked_from` key is returned.

Updating all the API to support track relinking:
- [x] Get a Track
- [x] Get Several Tracks
- [x] Get an Album
- [x] Get Several Albums
- [x] Get an Album’s Tracks
- [x] Get a Playlist’s Tracks
- [x] Get a User’s Saved Tracks

### DIFF
--- a/user.go
+++ b/user.go
@@ -100,11 +100,13 @@ func (c *Client) CurrentUsersShows() (*SavedShowPage, error) {
 
 // CurrentUsersShowsOpt is like CurrentUsersShows, but it accepts additional
 // options for sorting and filtering the results.
+// API Doc: https://developer.spotify.com/documentation/web-api/reference-beta/#endpoint-get-users-saved-shows
 func (c *Client) CurrentUsersShowsOpt(opt *Options) (*SavedShowPage, error) {
 	spotifyURL := c.baseURL + "me/shows"
 	if opt != nil {
 		v := url.Values{}
 		if opt.Country != nil {
+			// Note: This parameter doesn't actually exist
 			v.Set("country", *opt.Country)
 		}
 		if opt.Limit != nil {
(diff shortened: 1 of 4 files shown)


user.go
```diff
@@ -136,12 +138,13 @@ func (c *Client) CurrentUsersTracks() (*SavedTrackPage, error) {
...
 // CurrentUsersTracksOpt is like CurrentUsersTracks, but it accepts additional
 // options for sorting and filtering the results.
+// API Doc: https://developer.spotify.com/documentation/web-api/reference-beta/#endpoint-get-users-saved-tracks
 func (c *Client) CurrentUsersTracksOpt(opt *Options) (*SavedTrackPage, error) {
 	spotifyURL := c.baseURL + "me/tracks"
 	if opt != nil {
 		v := url.Values{}
 		if opt.Country != nil {
-			v.Set("country", *opt.Country)
+			v.Set("market", *opt.Country)
```
zmb3, line 147: Good catch - are there other places where we're using `country` and should switch to `market`?

zmb3: On second thought, I'm wondering if we should be reusing the Country option for this.  The country query parameter is more for filtering, while the market parameter does something entirely different.  

If we add support for track relinking, I would want to see all endpoints that support relinking updated.

fclairamb: > Good catch - are there other places where we're using `country` and should switch to `market`?

I didn't find any other in _this_ source code.

fclairamb:
Hi @zmb3, so what is the next step? Do you want me to change something or can we merge it as-is?

zmb3:
Sorry @fclairamb I didn't see your reply 😬

The web API docs show that the following calls support track relinking:

- Get a Track
- Get Several Tracks
- Get an Album
- Get Several Albums
- Get an Album’s Tracks
- Get a Playlist’s Tracks
- Get a User’s Saved Tracks

It looks like `"country"` is not valid for any of those, so I am fine overriding "country" in these cases to be used for track relinking, so long as the godoc makes that clear.

I do think it makes sense to get all of these calls supporting track relinking together though, rather than leave things in a state where relinking works for some things but not others. Are you up for making the same changes in all of these places?  I don't mind if it's part of the same PR or separate ones, so long as they all get updated in the same 1-2 days.

fclairamb:
Wowkay, I guess I can do that. I indeed only focused on fixing the parts where I needed this feature.

fclairamb:
Hi @zmb3, 
so I finally took the time to update the other parts of the library where the `market` parameter can be used.

Please note I broke the `GetAlbumTracksOpt` API because it's the only one of the `*Opt` function that didn't receive an `*Option` parameter. I think it's worth it here.

zmb3:
Thanks @fclairamb for taking the time, and happy new year!

fclairamb:
Great! Thank you and happy new year! 🎉

PR #142, opened 2020-12-31
Options syntax sugar proposal

fclairamb:
The idea is to vastly reduce the pain of using the `Options` while keeping the current API.

### DIFF
--- a/category_test.go
+++ b/category_test.go
@@ -90,12 +90,7 @@ func TestGetCategoryPlaylistsOpt(t *testing.T) {
 	})
 	defer server.Close()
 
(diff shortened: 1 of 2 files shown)


zmb3:
Love it.  Let's make sure one of the examples under `examples/` is using this (or add a new example) so that it's easily discoverable.

I'd also show a snippet using this approach in the README under the[ Optional Parameters](https://github.com/zmb3/spotify/blob/master/README.md#optional-parameters) section.

strideynet:
Closing as replaced with new options syntax.

PR #144, opened 2021-01-17
Proposed v2

strideynet:
### Small to-do list

- [x] Add linting and unit-testing on PRs
- [x] Add context.Context support where possible
- [ ] Functional options handling

# Changes

## context.Context

All operations that cross the network will now take a context.Context as their first parameter.

## Functional options handling

Move away from having a fixed struct and instead use functional options e.g https://github.com/strideynet/spotify-go/pull/1/commits/e381bd7cb91dd28019d1505951f191307071dd92 

This is more easily expanded in future and doesn't require odd behaviour from consuming code (e.g creating a integer for limit and immediately indirecting it as a psuedo-optional parameter)

## Linting and Testing

Linting and testing are now implemented using Github actions and run on commits to master and pull requests.


strideynet:
@zmb3 , I'd be particularly interested in your thoughts on my proposed interface for optional parameters demonstrated here: https://github.com/strideynet/spotify-go/commit/e381bd7cb91dd28019d1505951f191307071dd92

zmb3:
Love it. I will take a closer look at this later today. In the meantime, the v2 folder approach sounds like the right thing to do, and I’d be super happy to add you as a maintainer!

strideynet:
Perfect. I have a few thoughts/questions:

- Should we outright throw an error when a developer tries to use an option that is not compatible with an endpoint? As far as I can tell, Spotify ignores additional URL parameters so there's no harm other than potential user confusion.
- Are there any other outstanding changes that we should aim to incorporate into this v2 release?

zmb3:
> Should we outright throw an error when a developer tries to use an option that is not compatible with an endpoint? As far as I can tell, Spotify ignores additional URL parameters so there's no harm other than potential user confusion.

We've  (I've) been too strict with things in the past when Spotify didn't actually care, and that caused issues - so I would say if Spotify accepts the request and doesn't issue an error, then we shouldn't error either. 

> Are there any other outstanding changes that we should aim to incorporate into this v2 release?

I've never been super happy with the authenticator code. Wondering if we rewrite that or simply remove it and provide examples on how to use the standard OAuth packages to get a client/token. Thoughts?  There was also #107 which looked like a nice addition but wasn't backwards compatible so we decided not to merge.

.github/workflows/test.yml
```diff
@@ -0,0 +1,16 @@
...
+  pull_request:
+jobs:
+  unit:
+    name: unit
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '1.15.6' # The Go version to download (if necessary) and use.
```
zmb3, line 15: Does this action support something more flexible, like maybe just the latest 2 Go releases?  I always found it annoying to Go back and update a hard coded version when there's new patches or releases of Go.

album.go
```diff
@@ -136,13 +137,13 @@ func toStringSlice(ids []ID) []string {
...
-func (c *Client) GetAlbums(ids ...ID) ([]*FullAlbum, error) {
-	return c.GetAlbumsOpt(nil, ids...)
+func (c *Client) GetAlbums(ctx context.Context, ids ...ID) ([]*FullAlbum, error) {
+	return c.GetAlbumsOpt(ctx, nil, ids...)
 }
 
 // GetAlbumsOpt is like GetAlbums but it accepts an additional country option for track relinking
 // Doc API: https://developer.spotify.com/documentation/web-api/reference/albums/get-several-albums/
-func (c *Client) GetAlbumsOpt(opt *Options, ids ...ID) ([]*FullAlbum, error) {
+func (c *Client) GetAlbumsOpt(ctx context.Context, opt *Options, ids ...ID) ([]*FullAlbum, error) {
```
zmb3, line 146: Is the plan to replace `*Options` with the functional options pattern everywhere?

category.go
```diff
@@ -96,8 +97,8 @@ func (c *Client) GetCategoryPlaylistsOpt(catID string, opt *Options) (*SimplePla
 
 // GetCategories gets a list of categories used to tag items in Spotify
 // (on, for example, the Spotify player's "Browse" tab).
-func (c *Client) GetCategories() (*CategoryPage, error) {
-	return c.GetCategoriesOpt(nil, "")
+func (c *Client) GetCategories(ctx context.Context) (*CategoryPage, error) {
+	return c.GetCategoriesOpt(ctx,nil, "")
```
zmb3, line 101: Does this need to be `gofmt`'d?

options.go
```diff
@@ -0,0 +1,51 @@
...
+
+type Option func(*optionSet)
+
+type optionSet struct {
+	urlParams url.Values
+}
+
+// Limit sets the number of entries that a request should return
+func Limit(amount int) Option {
+	return func(oS *optionSet) {
```
zmb3, line 16: `oS` just looks weird to me.  How about `o` instead?

options.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Offset sets the index of the first entry to return
+func Offset(amount int) Option {
+	return func(oS *optionSet) {
+		oS.urlParams.Set("offset", strconv.Itoa(amount))
+	}
+}
+
+// Timerange sets t
```
zmb3, line 35: We need a better doc comment here.  What is `t`?

spotify_test.go
```diff
@@ -15,7 +16,7 @@ func testClient(code int, body io.Reader, validators ...func(*http.Request)) (*C
 			v(r)
 		}
 		w.WriteHeader(code)
-		io.Copy(w, body)
+		_, _ = io.Copy(w, body)
```
zmb3, line 19: Is a linter complaining about not checking the return values here?  I see little value in adding `_, _`.

zmb3 (commented):
Looking good.

It would be interesting to see if adopting functional options everywhere allows us to have a single version of each method, rather than a `Foo` and `FooOpt` variant.

strideynet:
I think I agree with the work in #107 meaning that library consumers would be more free to provide their own HTTP client/authentication, however, there seems to remain some value in keeping our auth component as that seems to work for most users.

> It would be interesting to see if adopting functional options everywhere allows us to have a single version of each method, rather than a Foo and FooOpt variant.

My intention (as shown in the example commit) was to rid us of the FooOpt variants, since we can just use variadic options instead, greatly simplifying the library/reducing code.

spotify_test.go
strideynet: Unfortunately yes, the linter doesn't like ignored error values. I could break from the suggested lint, but I tend to try to avoid this and stick to defaults where possible.

album.go
strideynet: Yup, that would be my plan if you are happy with the options as it is currently looking? I didn't want to go ahead and refactor everywhere until you seemed happy with it.

spotify_test.go
zmb3: Yeah, so let's just check the error instead of `_` it out.

zmb3:
Cool, I'm on board with keeping the authenticator then, so long as we have a way to also provide your own token or authenticated client should you choose to.

This is looking good. Looks like we still need to move things under a `/v2` folder. If you want to call this proposal accepted and start making the changes on a v2 branch, I've made you a maintainer.  Thanks for all the help!

.github/workflows/test.yml
strideynet: I can configure a matrix to run multiple go versions, but as far as I'm aware theres no support for latest or latest - 1

strideynet:
> Cool, I'm on board with keeping the authenticator then, so long as we have a way to also provide your own token or authenticated client should you choose to.
> 
> This is looking good. Looks like we still need to move things under a `/v2` folder. If you want to call this proposal accepted and start making the changes on a v2 branch, I've made you a maintainer. Thanks for all the help!

I'm actually not entirely sure that we have to use a separate folder, it may be possible to do it just with tagging as per https://stackoverflow.com/a/53380636 . This seems simpler to me if we have no intentions of hotfixing old releases.

zmb3:
Agree we have no intentions of supporting v1 when this work is complete.

I don't want to have to keep a v2 branch for v2 and let the master branch just sit idle. It looks like we can tag on master for v2, but I'm not entirely sure how that works. Hopefully we won't have to keep updating the v2 tag.

PR #145, opened 2021-01-17
V2

strideynet:
V2 should include the following breaking changes:

- [x] context.Context support
- [x] support for instantiation of the library with a custom HTTP client (without using our auth)
- [x] support for functional options rather than config structs for list endpoints


strideynet:
Hey @zmb3 , having some general thoughts that might be easier to discuss in IRC/Slack/Discord. Is there any platforms I can reach you on?

zmb3:
@strideynet you bet. I’m @zmb3 on Gophers slack.

zmb3:
@strideynet we should remove the AppVeyor CI as part of this work too. The Github actions are easier to maintain and we really don't do anything OS-specific so AppVeyor's Windows support is not adding vaue.

strideynet:
@zmb3 Yeah! I can't find the config for that, but Github Actions is much better integrated with github.

zmb3:
I deleted it from AppVeyor, so we just need to remove the badge from the README, and might as well delete the .travis.yml file too.

strideynet:
Perfect.

examples/authenticate/authcode/authenticate.go
```diff
@@ -58,8 +65,9 @@ func completeAuth(w http.ResponseWriter, r *http.Request) {
 		http.NotFound(w, r)
 		log.Fatalf("State mismatch: %s != %s\n", st, state)
 	}
+
 	// use the token to get an authenticated client
-	client := auth.NewClient(tok)
+	client := spotify.New(spotify.HTTPClientOpt(auth.Client(r.Context(), tok)))
```
strideynet, line 70: @zmb3 , would be interested to hear your thoughts on the new style here, with the spotify HTTPClientOpt being combined with the http client returned by Client() ?

zmb3: It looks good from a feature perspective, but if we expect that this is going to be the most common way to initialize a client I think we need to make it simpler. It's quite a mouthful the way it reads right now.

It's tough to review in this format with so many changes - what might help me visualize the API is a few examples of how you might create a new client (ie the simplest/default way, and then more verbose options if you need to provide your own token or HTTP client).

nzoschke:
Just wanted to say this is working well for me. I wanted to make a client that is wrapped in a local disk / memcache. I already have an access token handy from a saved OAuth flow so it looks like:

```
type CacheTransport struct {
	Dir       string
	Transport http.RoundTripper
}

func (t *CacheTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	h := sha1.New()
	h.Write([]byte(req.URL.String()))
	bs := h.Sum(nil)

	filename := filepath.Join(t.Dir, fmt.Sprintf("%x", bs))
	if fileExists(filename) {
		bs, err := ioutil.ReadFile(filename)
		if err != nil {
			return nil, err
		}

		return &http.Response{
			Body:       ioutil.NopCloser(bytes.NewBuffer(bs)),
			Status:     http.StatusText(http.StatusOK),
			StatusCode: http.StatusOK,
		}, nil
	}

	res, err := t.Transport.RoundTrip(req)
	if err != nil {
		return res, err
	}

	bs, err = ioutil.ReadAll(res.Body)
	if err != nil {
		return res, err
	}

	if err := ioutil.WriteFile(filename, bs, 0644); err != nil {
		return res, err
	}

	res.Body = ioutil.NopCloser(bytes.NewBuffer(bs))

	return res, nil
}

auth := spotifyauth.New("", "").Client(context.Background(), &oauth2.Token{
	AccessToken: tok.AccessToken,
	TokenType:   tok.TokenType,
})

tr, err := NewCacheTransport(auth.Transport)
if err != nil {
	return spotify.Client{}, xerrors.Errorf(": %w", err)
}
auth.Transport = tr

c := spotify.New(spotify.HTTPClientOpt(auth), spotify.RetryOpt(true))
```

I found this impossible to do with V1, so +1 on this restructuring.

strideynet:
Hey @nzoschke. Glad this is working great for you! Feel free to send any Qs or thoughts my way, really looking to get the dev experience sweet on this 👍

strideynet:
I've updated the Authenticator New() to take functional options as well. This stops the redundant empty string values when using Client Credentials or PKCE.

auth/auth.go
```diff
@@ -67,84 +67,107 @@ const (
...
+
+	return a
 }
 
-// SetAuthInfo overwrites the client ID and secret key used by the authenticator.
-// You can use this if you don't want to store this information in environment variables.
-func (a *Authenticator) SetAuthInfo(clientID, secretKey string) {
-	a.config.ClientID = clientID
-	a.config.ClientSecret = secretKey
+// WithHTTPClient returns a context with a value set to override the oauth2 http client as Spotify does not
```
zmb3, line 143:
```suggestion
// contextWithHTTPClient returns a context with a value set to override the oauth2 http client as Spotify does not
```

player.go
```diff
@@ -457,15 +438,16 @@ func (c *Client) PreviousOpt(opt *PlayOptions) error {
...
-	return c.SeekOpt(position, nil)
+func (c *Client) Seek(ctx context.Context, position int) error {
+	return c.SeekOpt(ctx, position, nil)
 }
 
 // SeekOpt is like Seek but with more options
 //
 // Only expects PlayOptions.DeviceID, all other options will be ignored
-func (c *Client) SeekOpt(position int, opt *PlayOptions) error {
+func (c *Client) SeekOpt(ctx context.Context, position int, opt *PlayOptions) error {
```
zmb3, line 448: Should we add functional options here too?

request_options.go
```diff
@@ -0,0 +1,122 @@
...
+//    fields = "tracks.items(track(name,href,album(!name,href)))"
+func Fields(fields string) RequestOption {
+	return func(o *requestOptions) {
+		o.urlParams.Set("fields", fields)
+	}
+}
+
+type Range string
+
+var (
```
Nightmarlin, line 96: Consider changing this to `const`?

strideynet: Good catch.

MalayCreates:
Hey all, I am having issues with GetTrack and GetTracks. I have tried to see where this could possibly come from. Could I get second hand confirmation that it is broken?

conradludgate:
I'd love to start using this in my project but I would like to wait for it to be merged first. Can I help in anyway?

strideynet:
@conradludgate, I'm actually hoping to do some work on this tomorrow. I'll let you know then if theres anything that comes to mind.

request_options.go
strideynet: Addressed.

player.go
strideynet: Good shout. I'll have a play with this now.

library.go
```diff
@@ -11,10 +11,20 @@ import (
...
+// UserHasAlbums checks if one or more albums are saved to the current user's
+// "Your Albums" library.
+func (c *Client) UserHasAlbums(ctx context.Context, ids ...ID) ([]bool, error) {
+	return c.libraryContains(ctx, "albums", ids...)
+}
+
+func (c *Client) libraryContains(ctx context.Context, typ string, ids ...ID) ([]bool, error) {
 	if l := len(ids); l == 0 || l > 50 {
-		return nil, errors.New("spotify: UserHasTracks supports 1 to 50 IDs per call")
+		return nil, errors.New("spotify: UserHasAlbums supports 1 to 50 IDs per call")
```
conradludgate, line 25: Ah my bad. This error is now wrong.
```suggestion
		return nil, errors.New("spotify: only supports 1 to 50 IDs per call")
```

player.go
strideynet: I'm not sure the same benefits apply here as they do with instantiating the client, so I'll move ahead without functional options on SeekOpt/PlayOpt for now

PR #146, opened 2021-01-25
Add ExchangeWithOpts to be used with pkce authentication

ialbors:
So we can set "code_verifier" param url without needing an
HTTP request.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -184,6 +184,12 @@ func (a Authenticator) Exchange(code string) (*oauth2.Token, error) {
 	return a.config.Exchange(a.context, code)
 }
 
+// ExchangeWithOpts performs the same function as the Exchange function
+// but takes in optional URL Auth params
+func (a Authenticator) ExchangeWithOpts(code string, opts ...oauth2.AuthCodeOption) (*oauth2.Token, error) {
+	return a.config.Exchange(a.context, code, opts...)
+}
+
 // NewClient creates a Client that will use the specified access token for its API requests.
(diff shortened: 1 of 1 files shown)


zmb3:
@strideynet is actually working on a v2 branch right now with better options support. Let's make sure to get this feature included there.

ialbors:
Is there anything I could do about it? 

Should I open a pull request against the other branch (if not already included)?

Thanks.

strideynet:
I can't see the harm in this particular change being made directly to master here as part of the 1.0 release series, and I'll merge master back into the v2 branch once that's been done :)

auth.go
```diff
@@ -184,6 +184,12 @@ func (a Authenticator) Exchange(code string) (*oauth2.Token, error) {
 	return a.config.Exchange(a.context, code)
 }
 
+// ExchangeWithOpts performs the same function as the Exchange function
+// but takes in optional URL Auth params
+func (a Authenticator) ExchangeWithOpts(code string, opts ...oauth2.AuthCodeOption) (*oauth2.Token, error) {
```
strideynet, line 189: Could we just add this variadic parameter to the main Exchange() method? It'll be non-breaking for existing users (since no options is valid) and keep the overall interface simpler.

strideynet (commented):
Looks good, just one thought

auth.go
ialbors: I'm ok with it.
I was just trying to mimic the behavior with Token and TokenWithOpts funcs.

Let me change it.

strideynet (approved):

Issue #147, opened 2021-02-14
panic: reflect: call of reflect.Value.Elem on struct Value


klokare:
I had a playlist with more than 100 tracks in it so I needed to go to the next page in the returned object. When calling client.NextPage(plist.Tracks), where plist is a *spotify.FullPlaylist, I received the following panic:
```
panic: reflect: call of reflect.Value.Elem on struct Value

goroutine 1 [running]:
reflect.Value.Elem(0x159f900, 0xc00009c0c0, 0x99, 0xc0000c6138, 0x60, 0xc0004c6000)
        /usr/local/go/src/reflect/value.go:843 +0x1a5
github.com/zmb3/spotify.(*Client).NextPage(0xc000591c80, 0x16a6800, 0xc00009c0c0, 0xc00009c0c0, 0x82)
        /Users/hummerb/.go/pkg/mod/github.com/zmb3/spotify@v0.0.0-20201231194903-e2d01d9b8bd2/page.go:105 +0xbe
jbox/data.EnsurePlaylist(0x16b12a0, 0xc00006cbc0, 0x16b4ba0, 0xc00016cea0, 0xc00009a480, 0x1604125, 0x1b, 0x1, 0x0, 0x0, ...)
```

The code in NextPage is expecting a reference. I changed the definition of FullPlaylist from 
```go
// FullPlaylist provides extra playlist data in addition to the data provided by SimplePlaylist.
type FullPlaylist struct {
	SimplePlaylist
	// The playlist description.  Only returned for modified, verified playlists.
	Description string `json:"description"`
	// Information about the followers of this playlist.
	Followers Followers          `json:"followers"`
	Tracks    PlaylistTrackPage `json:"tracks"`
}
```
to
```go
// FullPlaylist provides extra playlist data in addition to the data provided by SimplePlaylist.
type FullPlaylist struct {
	SimplePlaylist
	// The playlist description.  Only returned for modified, verified playlists.
	Description string `json:"description"`
	// Information about the followers of this playlist.
	Followers Followers          `json:"followers"`
	Tracks    *PlaylistTrackPage `json:"tracks"`
}
```

This corrected the problem for me.

strideynet:
Thanks for raising this. I'll take a look tonight :)

strideynet:
@klokare Can you please provide an in-context example of you invoking NextPage() ?

strideynet:
[...]

PR #148, opened 2021-02-15
Fix ambiguity in Paging

strideynet:
Fixes potential panic case without changing the API, by switching the page base struct to use a pointer receiver instead.

See #147  for more details


zmb3:
LGTM, @strideynet. Look forward to hearing your thoughts on paging in V2.

zmb3 (approved):

PR #154, opened 2021-03-29
add an example using paging object

RY908:
Added an example using Spotify's paging object.

### DIFF
--- a/examples/user/user.go
+++ b/examples/user/user.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


strideynet:
What is this demonstrating that isn't demonstrated in the paging demo? I'd prefer to try and keep the demo codebase as small as possible because its a lot of code to maintain. Perhaps it would make more sense to add to the existing paging demo if something is missing.

RY908:
Sorry for bothering. I didn't notice that there is already an example using paging object.

PR #163, opened 2021-08-13
feat: add library album apis

conradludgate:
Similar to how you can save tracks to your library, you can do the same with albums.

I copied the code directly from the tracks apis. It could probably be refactored to shared code a bit more


strideynet:
Thanks for this PR, I'll take a look at this tomorrow and test it

library_test.go
```diff
@@ -55,3 +55,54 @@ func TestRemoveTracksFromLibrary(t *testing.T) {
...
+
+func TestRemoveAlbumsFromLibrary(t *testing.T) {
+	client, server := testClientString(http.StatusOK, "")
+	defer server.Close()
+
+	err := client.RemoveAlbumsFromLibrary("4iV5W9uYEdYUVa79Axb7Rh", "1301WleyT98MSxVHPZCA6M")
+	if err != nil {
+		t.Error(err)
+	}
+}
```
strideynet, line 108: Just be picky, can I get a newline at the end of this file please

strideynet (changes requested):
PR looks good, just need a newline :)

library_test.go
conradludgate: Sure. It usually does it automatically but I was testing out codespaces :sweat_smile:

PR #164, opened 2021-08-15
fix: follow api

conradludgate:
Seems the follower api url is incorrect: https://developer.spotify.com/documentation/web-api/reference/#endpoint-follow-playlist
It does not need an 'owner id'


strideynet:
I'm avoiding breaking the API on V1, could you raise this as a PR against the V2 branch please ?

Also, if you are on the Gophers slack and want to chat about various spotify go library things, please feel free to DM me there :)

strideynet (approved):

PR #172, opened 2021-10-16
chore: add sample server side package usage

bretanac93:
This PR is intended to showcase how this library can be used without having to spin up a web server and instead directly use the auth-code provided by Spotify to negotiate an OAuth Token. This is specially useful when the first part of the process is done in a frontend and the backend only cares about the auth code the frontend provides.

### DIFF
--- a/examples/cli/cli.go
+++ b/examples/cli/cli.go
@@ -1,0 +1,39 @@
+package cli
+
+import (
+	"context"
+	"flag"
+	"github.com/zmb3/spotify/v2"
+	spotifyauth "github.com/zmb3/spotify/v2/auth"
+	"golang.org/x/oauth2"
+	"log"
+)
+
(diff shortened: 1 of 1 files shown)


examples/cli/cli.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"flag"
+	"github.com/zmb3/spotify/v2"
+	spotifyauth "github.com/zmb3/spotify/v2/auth"
+	"golang.org/x/oauth2"
+	"log"
+)
+
+var auth = spotifyauth.New(spotifyauth.WithRedirectURL("http://localhost:3000/login_check"))
```
strideynet, line 12: Do we need to set a redirect URL here ?

bretanac93: Yes, it is required by Spotify when exchanging the authorization code by the token for validation purposes (https://developer.spotify.com/documentation/general/guides/authorization/code-flow#request-access-token)

Issue #161, opened 2021-07-23
bug: PlaybackOffset with Position=0 doesn't work


insomnimus:
First I want to thank the maintainers of this package for their awesome work.

I've stumbled on this issue while working with my spotify shell, [libman](https://github.com/insomnimus/libman).

The snippet below does not work:

```go
client.PlayOpt(&spotify.PlayOptions{
  PlaybackContext: &someURI,
    PlaybackOffset: &spotify.PlaybackOffset{ Position: 0 },
    })
```

It says in the documentation that the position is 0 based, but since the field has `omitempty`, it will omit the field essentiall sending a bare PlaybackOffset.

Here's what spotify returns:

`error: At least one of "uri" or "position" should be specified`

_Edit: Fix md syntax._

strideynet:
Thanks for raising this. I'll try and take a look at this later.

omerhanci:
I guess this fix broke the usage of URI in PlaybackOffset. 

Following code returns: ` Illegal offset: should be position XOR uri`

```
opts.PlaybackOffset = &spotify.PlaybackOffset{
			URI: songUri,
		}
```

omerhanci:
Hi, I created a PR to fix this problem, can you check it? 
https://github.com/zmb3/spotify/pull/177

PR #177, opened 2021-12-05
PlaybackOffset Position field is updated from int to *int

omerhanci:
Hi, after the fix mentioned here: https://github.com/zmb3/spotify/issues/161 playing with the PlaybackOffset URI parameter seems to be broken. 

Although Position parameter is not set, since it has a default value 0 and there is no omitempty, it's always sent to spotify api. So that changing it to some nullable type and adding omitempty again seems to fix the issue here.


strideynet:
Hi @omerhanci, I'll review this tonight thank you.

strideynet:
This was resolved in another PR.

PR #179, opened 2021-12-27
auth: remove contextwithHTTPClient

StalkR:
Setting a custom Transport with no ability for the caller to configure
it breaks the ability to configure an http proxy, while the default
behavior via http.DefaultTransport is to support it (via env),
see https://pkg.go.dev/net/http#RoundTripper.

This was added because Spotify didn't support HTTP/2 in 2017, but it
does now. Also, Go net/http should auto negotiate better.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -2,7 +2,6 @@ package spotifyauth
 
 import (
 	"context"
-	"crypto/tls"
 	"errors"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


strideynet:
Thanks for raising this.

#20 documents the original implementation of this. (for my records)

I'll review this tonight potentially, as obviously this is quite a significant change and I just want to ensure this doesn't break anything.

strideynet:
Tested this locally. All seems to work nicely.

strideynet (approved):

PR #183, opened 2022-01-06
Remove 202 from shouldRetry

### LINKED ISSUE #182: Cannot Set Playlist Image With Autoretry
# Version
* Go: `go version go1.17.5 windows/amd64`
* `github.com/zmb3/spotify/v2`: `v2.0.1`

# Issue
When I invoke [`client.SetPlaylistImage`](https://github.com/zmb3/spotify/blob/v2.0.1/playlist.go#L542) **without** [`client.WithRetry`](https://github.com/zmb3/spotify/blob/v2.0.1/spotify.go#L56), everything works and the playlist image updates. When I invoke `client.SetPlaylistImage` **with** `client.WithRetry`, the playlist image is updated but `client.SetPlaylistImage` returns `"io: read/write on closed pipe"`.

# Expected Behavior
`client.SetPlaylistImage` should work the same with or without autoretry

# Possible Cause
The API endpoint ([`PUT /playlists/{playlist_id}/images`](https://developer.spotify.com/documentation/web-api/reference/#/operations/upload-custom-playlist-cover)) returns `202 Accepted` (Regardless if autoretry is enabled or not) which is then retried because of [`shouldRetry`](https://github.com/zmb3/spotify/blob/v2.0.1/spotify.go#L190). The reason that it works without autoretry is [this if statement](https://github.com/zmb3/spotify/blob/1543b5968ba7e16baac4782ce402c75815bd918c/spotify.go#L217). I'm not entirely sure if the problem is Spotify returning an unexpected response (Please note that this library seems to anticipate the 202 response [here](https://github.com/zmb3/spotify/blob/1543b5968ba7e16baac4782ce402c75815bd918c/playlist.go#L560)) or the condition in `shouldRetry`.

It seems this may have been discussed in #100 (Specifically [this](https://github.com/zmb3/spotify/pull/100#issuecomment-518732995) comment)

Wakeful-Cloud:
Fixes #182

**NOTE: I don't know if `http.StatusAccepted` is used by other endpoints in cases where the response needs to be retried.**

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -187,7 +187,7 @@ func (c *Client) decodeError(resp *http.Response) error {
 // shouldRetry determines whether the status code indicates that the
 // previous operation should be retried at a later time
 func shouldRetry(status int) bool {
-	return status == http.StatusAccepted || status == http.StatusTooManyRequests
+	return status == http.StatusTooManyRequests
 }
 
 // isFailure determines whether the code indicates failure


strideynet:
Thanks for raising this. I'll take a look tonight along with the other open PR.

strideynet:
I'm a little worried this might break some existing retries that are valid, So I'm going to need to take a bit of a deeper look.

Wakeful-Cloud:
@strideynet That was my concern too. I wonder if it's possible to analyze a response header/body to determine if the request actually failed or not. Are you aware of any known cases where Spotify returns 202 even though the request should be retried?

rtrox:
I was planning on addressing this too, then saw this PR. I think given that StatusAccepted was added for a reason, a safer approach might be to instead change the retry check to include a check of the needsStatus:

```go
if c.AutoRetry &&
   isFailure(resp.StatusCode, needsStatus) &&
   shouldRetry(resp.StatusCode) {
   // <retry logic>
}
```

rtrox:
gah, sorry, just realized I'm necro-ing a year old PR. will open a new PR for this.

Wakeful-Cloud:
#213 seems more complete, closing

PR #185, opened 2022-02-07
fixup: string types to ID, where it is spotify.ID + ID to string

jtagcat:


strideynet:
Hmm, unfortunately, changing the values of these types would be a breaking change, which means that'd we'd have to go to /v3. I'll have a think if there's a few changes we could bundle as part of this.

strideynet:
Reconsider this for the upcoming V3 milestone.

PR #189, opened 2022-05-20
Add GetPlaylistItems

xonstone:
For getting all tracks (both episode and track types) of a playlist

Web API ref: https://developer.spotify.com/documentation/web-api/reference/#/operations/get-playlists-tracks


playlist.go
```diff
@@ -190,6 +190,86 @@ func (c *Client) GetPlaylistTracks(
...
 
+// PlaylistItem contains info about an item in a playlist.
+type PlaylistItem struct {
+	// The date and time the track was added to the playlist.
+	// You can use the TimestampLayout constant to convert
+	// this field to a time.Time value.
+	// Warning: very old playlists may not populate this value.
+	AddedAt string `json:"added_at"`
+	// The Spotify user who added the track to the playlist.
+	// Warning: vary old playlists may not populate this value.
```
strideynet, line 201:
```suggestion
	// Warning: very old playlists may not populate this value.
```

playlist.go
```diff
@@ -190,6 +190,86 @@ func (c *Client) GetPlaylistTracks(
...
+	AddedBy User `json:"added_by"`
+	// Whether this track is a local file or not.
+	IsLocal bool `json:"is_local"`
+	// Information about the track.
+	Track PlaylistItemTrack `json:"track"`
+}
+
+// PlaylistItemTrack is a union type for both tracks and episodes. It defaults to track type.
+type PlaylistItemTrack struct {
+	*FullTrack
```
strideynet, line 211: I think it would be nicer if we just made this `Track *FullTrack`

strideynet (changes requested):
Nice work ! We should move towards suggesting users switch to this method from GetPlaylistTracks.

Can you:
- Add a deprecation comment to the GetPlaylistTracks method
- Include test data for a playlist with both a track and a episode
- Export the AdditionalTypes option, and fall back to providing both types if the user doesn't provide AdditionalTypes they desire?

playlist.go
xonstone: Changed it to full field.

xonstone:
I

> Nice work ! We should move towards suggesting users switch to this method from GetPlaylistTracks.
> 
> Can you:
> 
> * Add a deprecation comment to the GetPlaylistTracks method
> * Include test data for a playlist with both a track and a episode
> * Export the AdditionalTypes option, and fall back to providing both types if the user doesn't provide AdditionalTypes they desire? I essentially want this method to return both, unless the user explicitly states they just want one type.

I implemented all the mentioned changes as well.

playlist_test.go
```diff
@@ -141,6 +141,122 @@ func TestGetPlaylistTracks(t *testing.T) {
...
+	tm, err = time.Parse(TimestampLayout, added)
+	if err != nil {
+		t.Error(err)
+	}
+	if f := tm.Format(DateLayout); f != "2022-05-20" {
+		t.Errorf("Expected added at 2014-11-25, got %s\n", f)
+	}
+}
+
+func TestGetPlaylistItemsOverride(t *testing.T) {
```
strideynet, line 246: Can you add one test to check the default is both ?

strideynet (changes requested):
Looks good ! Just one last change please.

playlist_test.go
xonstone: Sure thing, added

strideynet (approved):

strideynet:
Really nice work, thank you <3

PR #190, opened 2022-05-21
Add ReplacePlaylistItems

xonstone:
Implementation of https://developer.spotify.com/documentation/web-api/reference/#/operations/reorder-or-replace-playlists-tracks replace functionality with URI's instead of trackID's

### DIFF
--- a/playlist_test.go
+++ b/playlist_test.go
@@ -560,6 +560,20 @@ func TestReplacePlaylistTracksForbidden(t *testing.T) {
 	}
 }
 
+func TestReplacePlaylistItemsForbidden(t *testing.T) {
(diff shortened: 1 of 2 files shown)


playlist_test.go
```diff
@@ -536,6 +536,35 @@ func TestReplacePlaylistTracks(t *testing.T) {
 	}
 }
 
+func TestReplacePlaylistItems(t *testing.T) {
```
strideynet, line 539: It would be nice if this was a table driven test for all scenarios in `ReplacePlaylistItems` rather than two seperate tests :)

xonstone: Replace the 2 separate tests with table tests

strideynet (approved):

PR #203, opened 2022-09-29
fix: description field in FullPlaylist was shadowing the one from SimplePlaylist

Sundava:
Follow up of #202, I realized that including the `Description` field in `SimplePlaylist` means it should removed from `FullPlaylist`.

It is very minor, but having it in both structs (In addition to being redundant) means that when explicitly selecting the `SimplePlaylist` part from the struct, the `Description` is always empty. It meant we couldn't do this :

```go
func printDescription(playlist SimplePlaylist) {
  fmt.Println(playlist.Description)
}

playlist, err := client.GetPlaylist(ctx, spotify.ID(spotifyID)) // Assume we fetch a playlist with a non empty Description
printDescription(playlist.SimplePlaylist) // Prints nothing, even if playlist.Description is not empty
}
```

### DIFF
--- a/test_data/get_playlist.txt
+++ b/test_data/get_playlist.txt
@@ -1,0 +1,460 @@
+{
+    "collaborative": false,
+    "description": "Bit of a overlap with phonk but whatever",
+    "external_urls": {
+        "spotify": "https://open.spotify.com/playlist/1h9q8vXXDl2vHOmwdsuXms"
+    },
+    "followers": {
+        "href": null,
+        "total": 0
+    },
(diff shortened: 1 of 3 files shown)


test_data/get_playlist.txt
```diff
@@ -0,0 +1,17856 @@
+{
```
strideynet, line 1: Is it possible to use a smaller playlist for this test data ?

strideynet (approved):

test_data/get_playlist.txt
Sundava: Sure thing
just pushed with a smaller one (~500 lines, without the extra line breaks in the `available_countries` field)

PR #212, opened 2022-12-28
Allow retry period to be cancelled by context

rtrox:
Currently, given the use of `time.Sleep`, retry time periods completely ignore the context passed into the function. This PR swaps the sleep logic to instead use a `select` and `time.After`, so that context cancellation will successfully cancel the sleep. The behavior in this PR is just to return the original error if context is cancelled during the retry Sleep. If you'd prefer, I can also swap to return a context cancellation, too.
Signed-off-by: Russell Troxel <russelltroxel@gmail.com>


strideynet:
Many thanks for this

PR #215, opened 2023-03-12
fix: support HTTP status 202

rmillet42:
even if spotify docs says its returns HTTP status 204, some endpoints return HTTP status 202


strideynet:
Classic Spotify API - Thanks for raising this @rmillet42

PR #220, opened 2023-04-12
Make spotify.PlaybackOffset.Position an *int 

### LINKED ISSUE #219: Illegal offset: should be position xor uri
Hi, I have a wrapper over the `client.PlayOpt` function that takes a context and a uri 
```go
func PlaySongWithContextURI(context, uri *spotify.URI) error {
	return Client.PlayOpt(ctx(), &spotify.PlayOptions{
		PlaybackContext: context,
		PlaybackOffset:  &spotify.PlaybackOffset{URI: uri},
	})
}
```
This function upon providing a context and uri gives the following error
```go
package main

import (
	"fmt"

	"github.com/zmb3/spotify/v2"
)

func main() {
	albumUri := spotify.URI("spotify:album:6E2VHYLPTtFOtdRcpt6YwP")
	trackUri := spotify.URI("spotify:track:4ME7UM3yALUZOxz5btBPc8")
	fmt.Println(PlaySongWithContextURI(&albumUri, &trackUri))
}

```

```bash
$ go run main.go 
Illegal offset: should be position XOR uri
```
The same request on spotify api dashboard runs perfectly fine

Apr12Wed065228AM

I think https://github.com/zmb3/spotify/commit/9d29507c9a13689c913905f76e69974e8b234072 introduced a bug where the position isn't omitted when empty. 
```go
bytes, err := json.Marshal(&PlaybackOptions{uri: "uri"})
fmt.Println(string(bytes))
// output: {"position":0,"uri":"uri"}
```
Upon changing back `PlaybackOffSet.Position` the error is resolved

```go
type PlaybackOffset struct {
	// Position is zero based and can’t be negative.
	Position int `json:"position,omitempty"`
	// URI is a string representing the uri of the item to start at.
	URI URI `json:"uri,omitempty"`
}
```
```bash
$ go run main.go
<nil>            
```
https://github.com/zmb3/spotify/commit/9d29507c9a13689c913905f76e69974e8b234072 removes the `omitempty` tag because the `Position` was ignored when it became zero. Using an integer pointer would fix this problem. As in

```go
type PlaybackOffset struct {
	// Position is zero based and can’t be negative.
	Position *int `json:"position,omitempty"`
	// URI is a string representing the uri of the item to start at.
	URI URI `json:"uri,omitempty"`
}
```
```bash
$ go run main.go
<nil>
```

aditya-K2:
Fixes #219 

https://github.com/zmb3/spotify/commit/9d29507c9a13689c913905f76e69974e8b234072 changed the JSON tag of the PlaybackOffset.Position resulting in it not being ignored when only URI was set. Which lead to the error mentioned #219. This PR fixes this by making PlaybackOffset.Position an integer pointer, hence it won't be ignored when it will be zero and will be ignored when only URI is set.

### DIFF
--- a/player.go
+++ b/player.go
@@ -84,7 +84,7 @@ type RecentlyPlayedResult struct {
 // the request may be accepted but with an unpredictable resulting action on playback.
 type PlaybackOffset struct {
 	// Position is zero based and can’t be negative.
-	Position int `json:"position"`
+	Position *int `json:"position,omitempty"`
 	// URI is a string representing the uri of the item to start at.
 	URI URI `json:"uri,omitempty"`
 }
--- a/player.go
+++ b/player.go
@@ -79,9 +79,13 @@ type RecentlyPlayedResult struct {
 	Items []RecentlyPlayedItem `json:"items"`
 }
 
-// PlaybackOffset can be specified either by track URI OR Position. If both are present the
-// request will return 400 BAD REQUEST. If incorrect values are provided for position or uri,
-// the request may be accepted but with an unpredictable resulting action on playback.
+// PlaybackOffset can be specified either by track URI OR Position. If the
+// Position field is set to a non-nil pointer, it will be taken into
+// consideration when specifying the playback offset. If the Position field is
+// set to a nil pointer, it will be ignored and only the URI will be used to
+// specify the offset. If both are present the request will return 400 BAD
+// REQUEST. If incorrect values are provided for position or uri, the request
+// may be accepted but with an unpredictable resulting action on playback.
 type PlaybackOffset struct {
 	// Position is zero based and can’t be negative.
 	Position *int `json:"position,omitempty"`


aditya-K2:
Any updates on this?

aditya-K2:
> This is a breaking change to the API of the client. I'll have a think if we want to merge this as-is and do a breaking release or if we'll need to come up with a different solution that still allows this to be controlled.

I understand that this is a breaking change to the API of the client, but I believe that the existing code is not functioning as expected, and this will only bring the library in line with the Spotify API specifications.

strideynet:
Yeah I agree, can you update the comment to explain what the behaviour is if this is nil and when a value is provided ?

aditya-K2:
> Yeah I agree, can you update the comment to explain what the behaviour is if this is nil and when a value is provided ?

Sure.

strideynet (approved):

Issue #232, opened 2023-07-15
Error getting current playing song: spotify: couldn't decode error: (46) [User not registered in the Developer Dashboard


TAMILVIP007:
https://nekobin.com/lujicuxepe.go hey there this is my complete oauth flow code

i am getting user id from user and trying to generate token and storing it to reuse it and get user current played songs. what i am stuck at is this error

`Error getting current playing song: spotify: couldn't decode error: (46) [User not registered in the Developer Dashboard`

i don't know why am i getting this error. It works fine initially when user try to get current track after that it returns this error. does it mean i cant re use token of other user with my client creds? or is there any other way to reuse or store any info which could get user track in each req instead of making them login again and again

TimHi:
Couple of things, you are only persisting and reusing the access token, if that is expired it will fail.
As mention in #221 the library will handle the case with an expired token, but the library needs the refresh token for that.

Your actual error message here indicates that a not registered user is trying to use your app. This is not related to this framework.

As described in the [dashboard documentation](https://developer.spotify.com/documentation/web-api/concepts/quota-modes) while you are in development mode with your spotify app you need to add the users manually.

TAMILVIP007:
how do i pass a refresh token?

TimHi:
I'm doing something like this in my project:
```
func createSpotifyClient(accessToken, refreshToken string, expiry time.Time, ctx context.Context) *spotify.Client {
	token := createToken(accessToken, refreshToken)
	httpClient := spotifyauth.New().Client(ctx, token)
	return spotify.New(httpClient)
}

func createToken(accessToken, refreshToken string) *oauth2.Token {
	token := &oauth2.Token{
		AccessToken:  accessToken,
		RefreshToken: refreshToken,
		TokenType:    "Bearer",
	}
	return token
}
```

In your code you're already doing something similiar, just extend this:
`	c, err := spotify.New(auth.Client(ctx.Background(), &oauth2.Token{AccessToken: token})).Token()
`

JunNishimura:
@TAMILVIP007 

Hi!
I solved this issue in the PR #233.

TAMILVIP007:
thanks a lot

Issue #221, opened 2023-04-16
refresh token


TAMILVIP007:
https://github.com/zmb3/spotify/blob/f897b41a6495bc134b79d4096d8d201f3970e344/examples/authenticate/authcode/authenticate.go#L31

hey there i used this method to generate oauth token to get user's current playing track. It shows it expires after one hour. How can i refresh token if its expired? without running a web auth again

cuotos:
this library is using the golang.org oauth2 client behind the scenes, it will handle the refresh for you when the token has expired.

"Client returns an HTTP client using the provided token. The token will auto-refresh as necessary. The underlying HTTP transport will be obtained using the provided context"

https://pkg.go.dev/golang.org/x/oauth2#Config.Client

TAMILVIP007:
Does it means i dont want to ask user to login each1h? after token gets expired?

ATroschke:
> this library is using the golang.org oauth2 client behind the scenes, it will handle the refresh for you when the token has expired.
> 
> "Client returns an HTTP client using the provided token. The token will auto-refresh as necessary. The underlying HTTP transport will be obtained using the provided context"
> 
> https://pkg.go.dev/golang.org/x/oauth2#Config.Client

I am sorry for waking this Thread up again, but is this true?

How can I hook into this?

Here is my current situation: I am asking users to connect their Spotify Account to my App to retrieve information about a User (specifically, their Followed Artists). I am then periodically synchronizing their Artists with a simple m2m Table in my DB and Fetching releases for my Users to display a proper and complete "New Releases" page.

I currently have an Issue where I frequently get revoked token Errors due to the refresh token already being used somewhere and had issues to track the exact place where it happens down.

ATroschke:
I'm just gonna answer myself here.

Spotify revokes refresh tokens after a single use, and the OAuth2 library did indeed automatically refresh my users access token.

PR #233, opened 2023-07-21
add RefreshToken method for Authenticator

JunNishimura:
## Overview
Existing code did not have a mechanism to update access tokens with refresh tokens.
So I added `RefreshToken` method which returns a new access token if it has expired.

## Example of use
```go
ctx := context.Background()
token := getToken() // get an existing token somehow
auth := spotifyauth.New()
newToken, err := auth.RefreshToken(ctx, token) // refresh token here
if err != nil {
  fmt.Println(err)
}
if token.AccessToken != newToken.AccessToken {
  saveToken(newToken) // save a new token
}
client := spotify.New(auth.Client(ctx, newToken))
```

## Related 
- #232 
- #221

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -171,8 +171,8 @@ func (a Authenticator) Token(ctx context.Context, state string, r *http.Request,
 	return a.config.Exchange(ctx, code, opts...)
 }
 
-// Return a new access token if it has expired.
-// If it has not expired, return the existing access token.
+// Return a new token if an access token has expired.
+// If it has not expired, return the existing token.
 func (a Authenticator) RefreshToken(ctx context.Context, token *oauth2.Token) (*oauth2.Token, error) {
 	src := a.config.TokenSource(ctx, token)
 	return src.Token()
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -171,6 +171,13 @@ func (a Authenticator) Token(ctx context.Context, state string, r *http.Request,
 	return a.config.Exchange(ctx, code, opts...)
 }
 
+// Return a new access token if it has expired.
+// If it has not expired, return the existing access token.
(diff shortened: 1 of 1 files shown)


JunNishimura:
@strideynet 
I added `RefreshToken` method.
I would appreciate it if you check this PR!

Padi142:
I would love to see this merged. I get a lot of problems with my tokens not refreshing

auth/auth.go
```diff
@@ -171,6 +171,13 @@ func (a Authenticator) Token(ctx context.Context, state string, r *http.Request,
 	return a.config.Exchange(ctx, code, opts...)
 }
 
+// Return a new token if an access token has expired.
+// If it has not expired, return the existing token.
+func (a Authenticator) RefreshToken(ctx context.Context, token *oauth2.Token) (*oauth2.Token, error) {
```
strideynet, line 176: I'm not sure calling this `RefreshToken` aligns with the GoDoc comment. Perhaps `GetToken` would be a clearer name - since the refreshing only occurs in certain circumstances.

JunNishimura: @strideynet 
Thank you for you opinion, and I understand what you are saying. 

One concern I have with naming it `GetToken` is that new users of this library may mistakenly think that they can get *oauth2.Token just by calling the `GetToken` function. (In fact, users need to call [Authenticator `Token` method](https://github.com/zmb3/spotify/blob/dc95f4def888572a00e6a3ae446d09b45cc914aa/auth/auth.go#L158-L172) to get the token for the first time.)

What do you think about this?

strideynet: I see what you are saying - and agree. Let's move forwards with this.

strideynet (approved):

Padi142:
<3 <3

PR #251, opened 2024-03-02
fix: unmarshal numeric fields regardless of them being int or float

### LINKED ISSUE #243: unmarshal fail.  【curl --request GET \   --url https://api.spotify.com/v1/artists/0TnOYISbd1XYRBk9myaseg】
json: cannot unmarshal number 1.1113535E7 into Go struct field Followers.followers.total of type uint

wernerdweight:
This fixes #243. I also updated the test data so that it contains float values.

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -102,60 +102,37 @@ func (id *ID) String() string {
 	return string(*id)
 }
 
+// Numeric is a convenience type for handling numbers sent as either integers or floats.
+type Numeric int
+
+// UnmarshalJSON unmarshals a JSON number (float or int) into the Numeric type.
+func (n *Numeric) UnmarshalJSON(data []byte) error {
+	var f float64
+	if err := json.Unmarshal(data, &f); err != nil {
+		return err
+	}
+	*n = Numeric(int(f))
+	return nil
+}
+
 // Followers contains information about the number of people following a
 // particular artist or playlist.
 type Followers struct {
 	// The total number of followers.
-	Count uint `json:"total"`
+	Count Numeric `json:"total"`
 	// A link to the Web API endpoint providing full details of the followers,
 	// or the empty string if this data is not available.
 	Endpoint string `json:"href"`
 }
 
-// UnmarshalJSON unmarshals the followers data regardless of numeric values being integers or floats.
-func (f *Followers) UnmarshalJSON(data []byte) error {
-	var v struct {
-		Count    float64 `json:"total"`
-		Endpoint string  `json:"href"`
-	}
-
-	if err := json.Unmarshal(data, &v); err != nil {
-		return err
-	}
-
-	f.Count = uint(v.Count)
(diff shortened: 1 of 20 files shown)


kreudom:
This does not work for me for the `FullTrack`. The fields stay zero.

I think the problem is here:
https://github.com/zmb3/spotify/blob/f5f7cf5e1671692f8dace79d61562e48d5d60914/track.go#L141-L148

Because `SimpleTrack` implements `UnmarshalJSON`, the struct `v` also implements it. Therefore `json.Unmarshal` directly calls `UnmarshalJSON` and ignores the other fields.

test_data/find_track.txt
strideynet, line 1: Could we duplicate the test data, rather than modifying the existing tests, to ensure we have a mixture of cases with floats and with integers, it looks like Spotify can't make their mind up and it'd be good to know we can handle both!

spotify.go
```diff
@@ -122,6 +139,25 @@ type Image struct {
...
+		Height float64 `json:"height"`
+		Width  float64 `json:"width"`
+		URL    string  `json:"url"`
+	}
+
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	i.Height = int(v.Height)
```
strideynet, line 154: I wonder if longterm it makes sense for us to return these fields as floats rather than casting them back to ints. It'd be a breaking change though so perhaps worth bundling with some other fixes.

strideynet:
Thanks for raising this - could you address my remark re: the test data ? I think it's important to prove we can handle both ints and floats. Then I'm happy to move forward with this.

test_data/find_track.txt
wernerdweight: done

spotify.go
wernerdweight: I kept it int for now, but removed the unmarshallers in favour of a custom type (more below)

wernerdweight:
@strideynet, I rewrote it with the custom type to avoid multiple unmarshallers. It's much simpler and flexible now. That said, it introduces a breaking change for comparisons. So if you, for some reason, compare one of the fields (e.g. the follower count) to an integer variable, it will break, as the custom type is not comparable to the int.
```go
playlist.Followers > 100 // still works
threshold := 100
playlist.Followers > threshold // doesn't work (mismatched types Numeric and int)
```

I still think this is a better solution than chaining the unmarshallers on top of each other, but since this is a bc-break, we/you could consider switching to floats (which would actually be a small change now that there's a custom type). Thanks

strideynet:
> I still think this is a better solution than chaining the unmarshallers on top of each other, but since this is a bc-break, we/you could consider switching to floats (which would actually be a small change now that there's a custom type).

I wonder if just switching to floats may make more sense if that's the type the Spotify API is now returning. We could cut this as a breaking change release.

Issue #257, opened 2024-03-29
client.Search for SearchTypeShow returns zero values


Santiago-Balcero:
Hi!

Currently working in a project with this Spotify client. When using client.Search for SearchTypeShow it always returns zero values of type SimpleShowPage.

I create client in this way:
```go
ctx := context.Background()
	authConfig := &clientcredentials.Config{
		ClientID:     SpotifyId,
		ClientSecret: SpotifySecret,
		TokenURL:     spotifyauth.TokenURL,
	}

	accessToken, err := authConfig.Token(ctx)
	if err != nil {
		log.Fatal("Error creating spotify client: ", err)
	}
	httpClient := spotifyauth.New().Client(ctx, accessToken)
	Spotify = *spotify.New(httpClient)
```

This is the line I use to call the search method, being `podcastName` a string:
```go
result, err := client.Search(ctx, podcastName, spotify.SearchTypeShow)
if err != nil {
	return "", fmt.Errorf("error searching podcast: %v", err)
}

log.Println(result.Shows)
```

And this is what I get as `result.Shows` no matter which `podcastName` value I use:
```bash
[...]

PR #258, opened 2024-04-14
fix: force market as request option to search for shows

Santiago-Balcero:
As put in this issue [https://github.com/zmb3/spotify/issues/257](https://github.com/zmb3/spotify/issues/257) calling for `client.Search` with `SearchTypeShow` was always returning zero values for `Shows` data.

When doing same request I got same issue, so when looking into Spotify Developer Forums [here](https://community.spotify.com/t5/Spotify-for-Developers/Search-for-shows-returns-an-array-of-null/m-p/4953378) and [here](https://community.spotify.com/t5/Spotify-for-Developers/Searching-for-podcasts-not-returning-results/m-p/4973209) it turned out to be a Spotify unfixed issue since, at least, 2020.

The solution to this as put by the community is to add market as query parameter as in:
```bash
curl --location 'https://api.spotify.com/v1/search?q=cyclast&type=show&market=co' \
--header 'Authorization: Bearer your-token'
```

This PR has:
- Enforces the use of market parameter when using `client.Search` with `SearchTypeShow` so `client` won't return zero values for correct search values.
- Creates a unit test for `client.Search` with `SearchTypeShow`.
- Adds `CountryColombia`.


search.go
```diff
@@ -124,6 +125,10 @@ func (c *Client) Search(ctx context.Context, query string, t SearchType, opts ..
 	v.Set("q", query)
 	v.Set("type", t.encode())
 
+	if t == SearchTypeShow && v.Get("market") == "" {
+		return nil, fmt.Errorf("searching for shows requires market parameter")
```
subash774, line 129: maybe worth creating a new error type if it doesn't already exist like 
```go
var ErrSearchMarketRequired = errors.New("spotify::search: searching for shows requires market parameter")
```

search.go
```diff
@@ -68,14 +69,14 @@ type SearchResult struct {
 // search types.  For example, `Search(query, SearchTypeArtist|SearchTypeAlbum)`
 // will search for artists or albums matching the specified keywords.
 //
-// Matching
+// # Matching
```
subash774, line 72: not sure if markdown style comments add any value 🤔

subash774:
after a quick local test session, I don't think market is the core issue here, when using `curl` or `postman` I can get the results without having to pass in the `market` parameter, something else is wrong, probably the way we're parsing the shows, will have a further look into it later
image

subash774:
aplogies, I was wrong, found the issue - when using bearer token generated from developer.spotify.com, the returned json contains all the data we need, however, when we use the client id and secret to generate a bearer token on each request, using the token spotify returns does not return any shows unless `market` parameter is passed in. Issue is on spotify side, @zmb3 @strideynet not sure how you wanna proceed here

search.go
Santiago-Balcero: Agree with this and will include it as part of PR.

search.go
Santiago-Balcero: My bad on this one. Will remove the extra #s.

Santiago-Balcero:
Followed @subash774 suggestion on new Error. As seen on Spotify forums it is on their side. Waiting for @zmb3 and @strideynet answer to see if they will proceed with this PR or not. Anyways I will be glad to help.

search.go
```diff
@@ -124,6 +129,10 @@ func (c *Client) Search(ctx context.Context, query string, t SearchType, opts ..
 	v.Set("q", query)
 	v.Set("type", t.encode())
 
+	if t == SearchTypeShow && v.Get("market") == "" {
```
strideynet, line 132: I'm not sure this is specific to searching Shows - but I'm also not sure that returning an Error here is the right thing to do. Spotify could adjust their API behaviour in future. Perhaps we just need to better document that when using Client ID/Secret you need to specify the market as there's no default market.

subash774: agreed, better documentation is probably the way to go as we don't control the scope of these params or if/when this issue really will get fixed. we can't decode the bearer token to get user info for us to implement the fallback method can we?

strideynet: > we can't decode the bearer token to get user info for us to implement the fallback method can we?

I think we'd probably want to avoid that - even if the bearer token was decodable it'd probably be a bad idea for us to couple with that.

Issue #171, opened 2021-08-22
Make Previous/Next page more mockable


strideynet:
As it stands, the pageable interface is not exported, which makes life more difficult mocking this.

emilbreiner99:
Hey there @strideynet! I ran into this while implementing some tests for my current project. Because the pageable interface is not exported but exposed in `NextPage(ctx context.Context, p pageable) error`, mocking is pretty tough and gets me into weird places (at least with my current go skills, learning it right now). Is there any not so obvious reason for the interface not being exported? I took a quick look, exported it and tested it in my current project. Worked like a charm. Because of this I and the low effort I already opened a PR for this in #260.

PR #260, opened 2024-05-20
Export pageable interface to ease up mocking

ebreiner:
The pageable interface was not exported but exposed in the client when retrieving the next page of pageable. This made mocking pretty tough.


emilbreiner99:
Closes #171

strideynet:
Does this actually make matters any better ? IIRC an interface with a unexported method cannot be implemented by another package. We'd also need to make `canPage` exported.

PR #270, opened 2024-11-13
Remove the Version constant

flimzy:
~It may be a better idea simply to remove this constant, as it's not referenced anywhere internally, and clearly hasn't been updated in long time, so it seems unlikely anybody is using it for anything valuable.  And v3 is a nice time to remove unused things.~

Remove woefully outdated Version constant

### DIFF
--- a/spotify.go
+++ b/spotify.go
@@ -17,9 +17,6 @@ import (
 	"golang.org/x/oauth2"
 )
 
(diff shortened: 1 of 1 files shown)


strideynet (approved):
👍  I'm happy to see this removed if you'd rather do that. I don't think it provides any value.

PR #273, opened 2024-11-14
added maxRetryDuration option to client

Pineapple217:
See #263

PR #274, opened 2024-11-15
Expose RetryAfter value in error

flimzy:
Related to, and useful along with #273

### DIFF
--- a/spotify_test.go
+++ b/spotify_test.go
@@ -2,14 +2,17 @@ package spotify
 
 import (
 	"context"
-	"golang.org/x/oauth2"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
+
+	"golang.org/x/oauth2"
 )
 
 func testClient(code int, body io.Reader, validators ...func(*http.Request)) (*Client, *httptest.Server) {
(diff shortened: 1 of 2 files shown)


Pineapple217:
IMO the Error type should contain time.Duration instead of time.Time. If the user wants time.Time they can calculate it them self (or add a method to Error to calculate it on the fly if you want it in the lib). Because if you want time.Duration you are just calculating back the value (doing double work). I know its nitpicky but something to think about. Either way I would be happy with this feature.

flimzy:
> IMO the Error type should contain time.Duration instead of time.Time.

That was my initial thought as well. The problem with a `time.Duration` is that it looses its meaning virtually immediately. If you're doing any sort of batch processing, or anything else that might delay the handling of an error, then that value cannot be relied on any more.  By returning a `time.Time`, we aren't forcing the consumer of this library into any specific error handling behavior or pattern.

Pineapple217:
Fair enough, both have their pros and cons but `time.Time` is probably better fitting for this lib when you put it like that. More intuitive and flexible.

spotify_test.go
```diff
@@ -104,6 +107,52 @@ func TestNewReleasesRateLimitExceeded(t *testing.T) {
 	}
 }
 
+func TestRateLimitExceededReportsRetryAfter(t *testing.T) {
+	t.Parallel()
+	const retryAfter = 2
+
+	handlers := []http.HandlerFunc{
```
strideynet, line 114: I've actually never thought about using a slice over a set of a funcs like this in a test - but this is quite nice!

strideynet (approved):

PR #284, opened 2025-04-16
More consistently include the HTTP response code in errors

flimzy:
Today's Spotify outage, and the plethora of random errors it's causing my service to return, inspired me to make errors a bit more parseable.

This PR does two things:

- It adds a convenience `HTTPStatus() int` method on the custom `spotify.Error` type, as an easy way to convient the HTTP status.
- It ensures that all errors actually include said HTTP status, when it's available.

It also changes the wording of some errors, for consistency, which may break some consumers of this library, if they depend on error string matching, which this PR will help make less necessary.

### DIFF
--- a/spotify_test.go
+++ b/spotify_test.go
@@ -115,7 +115,7 @@ func TestRateLimitExceededReportsRetryAfter(t *testing.T) {
 		// first attempt fails
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Set("Retry-After", strconv.Itoa(retryAfter))
-			w.WriteHeader(rateLimitExceededStatusCode)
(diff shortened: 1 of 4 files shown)


spotify_test.go
```diff
@@ -222,7 +222,7 @@ func TestDecode429Error(t *testing.T) {
 	if err == nil {
 		t.Fatal("Expected error")
 	}
-	if err.Error() != "Too many requests" {
-		t.Error("Invalid error message:", err.Error())
+	if err.Error() != "spotify: Too many requests [429]" {
+		t.Error("Unexpected error message:", err.Error())
```
strideynet, line 226: Nit: fairly trivial, but it might be nice to see an assertion here that `HTTPStatus` returns 429 as well?

